PR #7, opened 2015-09-01
contrib: CERN<->group linking

crepererum:
* Links user to existing groups if invenio-groups is installed.

Signed-off-by: Marco Neumann <marco@crepererum.net>

### DIFF
--- a/invenio_oauthclient/contrib/cern.py
+++ b/invenio_oauthclient/contrib/cern.py
@@ -93,6 +93,8 @@ from flask_login import current_user
 
 import requests
 
+from sqlalchemy.exc import IntegrityError
+
 #: Tunable list of groups to be hidden.
 CFG_EXTERNAL_AUTH_HIDDEN_GROUPS = (
     'All Exchange People',
@@ -209,16 +211,68 @@ def account_setup(remote, token):
 
     response = remote.get(REMOTE_APP_RESOURCE_API_URL)
     user = token.remote_account.user
+    if not user:
+        raise Exception('`user` should not be None, something bad happened!')
 
     if response.status == requests.codes.ok:
         res = get_dict_from_response(response)
-        current_user.info['group'] = fetch_groups(res['Group'])
+
+        # update session object
+        groups_old = current_user.info['group']
+        groups_new = fetch_groups(res['Group'])
+        current_user.info['group'] = groups_new
         current_user.modified = True
         current_user.save()
 
-        if user and not any([user.family_name, user.given_names]):
+        # update DB object
+        modified = False
+
+        # try to add/remove groups
+        try:
+            from invenio_groups.models import Group
+
+            # create sets to cut from O(n^2) to O(n * log(n))
+            groups_old_set = set(groups_old)
+            groups_new_set = set(groups_new)
+
+            for groupname in groups_new_set - groups_old_set:
+                candidates = Group.query \
+                    .filter(Group.name == groupname) \
+                    .filter(Group.is_managed == True) \
+                    .all()
+                for candidate in candidates:
+                    try:
+                        candidate.add_member(user)
+                        db.session.add(candidate)
+                        modified = True
+                    except IntegrityError:
+                        # user is already member of this group
+                        # => we are fine
+                        pass
+            for groupname in groups_old_set - groups_new_set:
+                candidates = Group.query \
+                    .filter(Group.name == groupname) \
+                    .filter(Group.is_managed == True) \
+                    .all()
+                for candidate in candidates:
+                    try:
+                        candidate.remove_member(user)
+                        db.session.add(candidate)
+                        modified = True
+                    except IntegrityError:
+                        # user is not a member of this group
+                        # => we are fine
+                        pass
+        except ImportError:
+            # no invenio-groups installed
+            pass
+
+        # try to add firstname+lastname if not set
+        if not any([user.family_name, user.given_names]):
             user.family_name = res['Lastname'][0]
             user.given_names = res['Firstname'][0]
-
             db.session.add(user)
+            modified = True
+
+        if modified:
             current_user.reload()


invenio_oauthclient/contrib/cern.py
```diff
@@ -212,13 +212,34 @@ def account_setup(remote, token):
...
         current_user.save()
 
-        if user and not any([user.family_name, user.given_names]):
-            user.family_name = res['Lastname'][0]
-            user.given_names = res['Firstname'][0]
-
-            db.session.add(user)
-            current_user.reload()
+        # update user object
+        if user:
```
jirikuncar: is it possible that we don't get a user from https://github.com/inveniosoftware/invenio-oauthclient/pull/7/files#diff-b49dd27458efb60f2de74978e6d3eecfR211?

crepererum: CC @ludmilamarian what do you think?

ludmilamarian: I'm thinking that the cern client (as well as the other clients) should only be a thin layer that handles specific client data, and all the logic and making sure that the user/token/etc. are well created and made available to the application should be taken care of in the oauthclient handlers - which I think is already the case. account_setup function should not even be called if anything is wrong with creating/fetching the user.

invenio_oauthclient/contrib/cern.py
```diff
@@ -212,13 +212,34 @@ def account_setup(remote, token):
...
+            try:
+                from invenio_groups.models import Group
+                for groupname in groups:
+                    candidates = Group.query.filter(Group.name == groupname).all()
+                    for candidate in candidates:
+                        candidate.add_member(user)
+                        db.session.add(candidate)
+                        modified = True
+            except ImportError:
+                # no invenio-groups installed
```
ludmilamarian: Can you register a warning in this case, so that it's clear for the admin why groups are not updated? 
I'm not sure how we will use the invenio-groups module - I don't think we are using the groups now, but it's good to have it in the logs if anything is expected, but it's not installed.
One more thing: how do you remove people from groups in this case?

ludmilamarian: This is what I was discussing with @crepererum IRL just now - we are not using Invenio groups as far as I know, but we are relying on the groups that we receive each time via SSO and are stored in the user_info / current_user object. So we don;t really need any synchro.

invenio_oauthclient/contrib/cern.py
```diff
@@ -209,16 +211,68 @@ def account_setup(remote, token):
...
+        current_user.info['group'] = groups_new
         current_user.modified = True
         current_user.save()
 
-        if user and not any([user.family_name, user.given_names]):
+        # update DB object
+        modified = False
+
+        # try to add/remove groups
+        try:
```
lnielsen: Wouldn't it be better to complement this via a signal?

crepererum: It is there to decide if we do a `current_user.reload()`. Ideally, our `invenio.ext.login.legacy_user` module should know do this on its own.

invenio_oauthclient/contrib/cern.py
```diff
@@ -209,16 +211,68 @@ def account_setup(remote, token):
...
+        raise Exception('`user` should not be None, something bad happened!')
 
     if response.status == requests.codes.ok:
         res = get_dict_from_response(response)
-        current_user.info['group'] = fetch_groups(res['Group'])
+
+        # update session object
+        groups_old = current_user.info['group']
+        groups_new = fetch_groups(res['Group'])
+        current_user.info['group'] = groups_new
```
jirikuncar: I'm not sure if we really want to maintain a copy of CERN group database in Invenio. IMHO it should be enough to `current_user.info['group'].append( fetch_groups(res['Group']))`.

--
cc @ludmilamarian

jirikuncar: Another thing is that the `invenio_oauthclient/contrib/cern` could register to `after-user-login` signal to extend `group` key in `current_user`.

kaplun: The original usage of having a copy of CERN egroups into Invenio was that then it was possible to do e.g. messaging/notifications to groups, independently to the single user being authenticated.

crepererum: Furthermore, groups can have descriptions and (not implemented yet) information pages. This gives users the ability to actually see (for some groups) which groups they are in and to interact with them.

ludmilamarian: I don't know if Invenio is the right place for CERN people to interact with their CERN egroups:
(1) we could not offer more than a read access to the list of the user's e-groups (which we can do without storing them into invenio-groups)
(2) there are several other places where a person can view their groups at CERN, more strongly related with the user's CERN account.

@kaplun I see that usergroup data is heavily used in baskets, but not much in other places. Do you know other use cases? Even if the user is not logged in, why do we need more than just the name of the group to do notifications? Also, I don't think in the legacy code that we are doing any removal of users from e-groups or purging of the e-groups that have been deleted, so we are just accumulating information, which is not that nice.

kaplun: In legacy we did purge users, by regularly polling e-groups information: https://github.com/inveniosoftware/invenio/blob/legacy/modules/websession/lib/webgroup.py#L831

On the other hand empty e-groups were never deleted. Correct.

In fact if nowadays CERN e-groups offer an easy runtime API to fetch membership information, then there is no point to replicate e-group in Invenio.

P.s. other places where e-group were highly used were authorizations and in theory WebMessage :smile:

ludmilamarian: @kaplun authorizations are using groups from the usergroup table? Isn't only firewall rules with "harcoded" e-group names and the logged in user's egroups from account_info (that we fetch via sso at login time)?

kaplun: @ludmilamarian, correct. Upon login time, external e-groups are synchronized into the usergroup table, then any usergroup (external or internal) from the usergroup table is loaded into `user_info` and it's then used for checking authorizations of that given user.

There is also a last use-case though:
https://github.com/inveniosoftware/invenio/blob/legacy/modules/webaccess/lib/access_control_firerole.py#L174

where the function `acc_firerole_extract_emails()` has a special `if CFG_CERN_SITE` to actually use external e-groups from the usergroup table, which is then used by `acc_get_authorized_emails()`. Which finally is only used in the infamous `publiline.py` monster, and in `webcomment`. In order to discover who should really be notified to moderate comments (even considering fireroles)

jirikuncar:
Since #19 has been merged, one can use new signals to update groups.

PR #9, opened 2015-09-11
installation: fix invenio_base imports

SamiHiltunen:
* FIX Adds missing `invenio_base` dependency and amends past
  upgrade recipes following its separation into standalone package.

Signed-off-by: Sami Hiltunen <sami.mikael.hiltunen@cern.ch>

### DIFF
--- a/invenio_oauthclient/forms.py
+++ b/invenio_oauthclient/forms.py
@@ -19,7 +19,7 @@
 
 """Forms for module."""
 
-from invenio.base.i18n import _
+from invenio_base.i18n import _
 from invenio_accounts.models import User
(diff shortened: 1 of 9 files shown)


invenio_oauthclient/config.py
```diff
@@ -122,7 +122,7 @@
     )
 
 The ``app_key`` parameter allows you to put your sensitive client id and secret
-in your instance configuration (``var/invenio.base-instance/invenio.cfg``).
+in your instance configuration (``var/invenio_base-instance/invenio.cfg``).
```
jirikuncar: please revert this change

PR #36, opened 2016-01-21
WIP contrib: CERN plugin fixes

egabancho:
cc @pamfilos @JavierDelgadoFernandez

### DIFF
--- a/examples/cern_app.py
+++ b/examples/cern_app.py
@@ -124,17 +124,12 @@ CERN_APP_CREDENTIALS = dict(
 # Create Flask application
 app = Flask(__name__)
 
-CERN_REMOTE_APP = copy.deepcopy(cern.REMOTE_APP)
-CERN_REMOTE_APP["params"].update(dict(request_token_params={
-    "scope": "Name Email Bio Groups",
-}))
-
 app.config.update(
     SQLALCHEMY_DATABASE_URI=os.environ.get(
         'SQLALCHEMY_DATABASE_URI', 'sqlite:///cern_app.db'
     ),
     OAUTHCLIENT_REMOTE_APPS=dict(
-        cern=CERN_REMOTE_APP
+        cern=cern.REMOTE_APP
     ),
     CERN_APP_CREDENTIALS=CERN_APP_CREDENTIALS,
     DEBUG=True,
@@ -158,6 +153,7 @@ app.register_blueprint(blueprint_user)
 app.register_blueprint(blueprint_client)
 app.register_blueprint(blueprint_settings)
 principal = app.extensions['security'].principal
+
 
 # FIXME: This should probably go into invenio-accounts
 @principal.identity_loader
--- a/examples/cern_app.py
+++ b/examples/cern_app.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2015 CERN.
+# Copyright (C) 2015, 2016 CERN.
 #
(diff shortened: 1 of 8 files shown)


examples/cern_app.py
```diff
@@ -142,10 +152,30 @@
...
+        return identity
+    except KeyError:
+        return None
+
+
+@principal.identity_saver
+def identity_saver_session(identity):
+    session['identity.id'] = identity.id
+    session['identity.auth_type'] = identity.auth_type
+    session['identity.provides'] = list(identity.provides)
```
jirikuncar: `s/list/set/`?

invenio_oauthclient/contrib/cern.py
```diff
@@ -190,15 +194,38 @@ def account_info(remote, resp):
...
 
     email = res['EmailAddress'][0]
-    common_name = res['CommonName'][0]
+    external_id = res['PersonID'][0]
+    # nice = res['CommonName'][0]
+    # name = res['DisplayName'][0]
 
     # FIXME get user informations accordly with invenio-userprofiles
-    return dict(email=email.lower(), nickname=common_name)
+    return dict(email=email.lower(),
```
jirikuncar: If `invenio-userprofiles` is installed you can add `profile=dict(nickname=nice, ...)`

invenio_oauthclient/contrib/cern.py
```diff
@@ -190,15 +194,38 @@ def account_info(remote, resp):
...
+        external_id = res['PersonID'][0]
+
+        # Create user <-> external id link.
+        oauth_link_external_id(user, dict(id=external_id, method="cern"))
+
+    groups = res['Group']
+    for group in groups:
+        g.identity.provides.add(RoleNeed(group))
+
+    identity_changed.send(current_app._get_current_object(),
```
jirikuncar: I'm not sure that you should send this signal.

egabancho: either I do that or I call directly `Principal.set_identity(identity)` https://github.com/mattupstate/flask-principal/blob/master/flask_principal.py#L412 (which will have more or less the same effect), otherwise the changes that I'm making to the identity will not be saved to the session and therefore lost

jirikuncar: If this is a problem we should just use `set_identity` as you proposed (https://github.com/mattupstate/flask-principal/blob/master/flask_principal.py#L412-L420)

egabancho: :+1:

invenio_oauthclient/contrib/cern.py
```diff
@@ -190,15 +194,39 @@ def account_info(remote, resp):
     """Retrieve remote account information used to find local user."""
     # Query CERN Resources to get user info and groups
     response = remote.get(REMOTE_APP_RESOURCE_API_URL)
+    current_app.logger.error(response.data)
```
jirikuncar: !

egabancho: Don't be impatient :wink:

jirikuncar: It's just a note :smiling_imp:

invenio_oauthclient/utils.py
```diff
@@ -96,6 +96,7 @@ def oauth_register(account_info, form_data=None):
         email = form_data.get("email")
 
     if email:
+        form_data = form_data or {'email': email}
```
jirikuncar: @egabancho can you add a test-case to cover this fix?

PR #37, opened 2016-01-25
tests: MySQL and PostgreSQL support

SamiHiltunen:
* BETTER MySQL and PostgreSQL are now supported.

* Enables MySQL and PostgreSQL testing on Travis.

Signed-off-by: Sami Hiltunen <sami.mikael.hiltunen@cern.ch>


jirikuncar:
@SamiHiltunen do we still need `TextEncryptedType` or can we remove it as we have done in Invenio-OAuth2Server?

PR #61, opened 2016-06-21
global: compatibility fix for OAuthlib>=0.9.0

jirikuncar:


lnielsen:
integration needs to be coordinated with Invenio-OAuth2Server to prevent version conflict:
https://github.com/inveniosoftware/invenio-oauth2server/blob/master/setup.py#L81

jirikuncar:
>  Invenio-OAuth2Server to prevent version conflict:

I'm working on it, therefore WIP ...

PR #64, opened 2016-06-23
contrib: minor CERN plugin fix

omelkonian:
* Fixes an issue where `g.identity.provides` was populated with only the
 	 Cern groups. As access rights can be assigned to single users(i.e. emails),
 	 the user's e-mail must also be included in the `g.identity.provides`.

### DIFF
--- a/invenio_oauthclient/contrib/cern.py
+++ b/invenio_oauthclient/contrib/cern.py
@@ -87,7 +87,7 @@ import copy
 import re
 
 from flask import current_app, session
-from flask_principal import RoleNeed, identity_loaded
+from flask_principal import UserNeed, RoleNeed, identity_loaded
 
 from invenio_oauthclient.utils import oauth_link_external_id
 from invenio_db import db
@@ -236,7 +236,9 @@ def account_setup(remote, token, resp):
         oauth_link_external_id(user, dict(id=external_id, method='cern'))
 
     groups = fetch_groups(res['Group'])
-    session['identity.cern_provides'] = [RoleNeed(group) for group in groups]
+    provides = [UserNeed(user.email)] + \
+               [RoleNeed('{0}@cern.ch'.format(group)) for group in groups]
+    session['identity.cern_provides'] = provides
 
 
 @identity_loaded.connect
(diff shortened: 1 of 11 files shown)


jirikuncar:
@omelkonian please amend commit message (`s/Cern/CERN/`) and reformat the bullet points (`kwalitee check message`).

omelkonian:
Although I cannot find any occurence of 'Cern'.

jirikuncar:
@omelkonian look to the commit message ...

jirikuncar:
@omelkonian please remove merge commit.

jirikuncar:
@omelkonian I have checked the diff again and I see that you have replaced the quotes everywhere. It's ok, but it should be in separate commit. Also make sure that you write CERN and not Cern.

jirikuncar:
WDYT about `s/contrib: strings' formatting/global: quotes usage unification/`?

invenio_oauthclient/contrib/cern.py
```diff
@@ -233,10 +233,12 @@ def account_setup(remote, token, resp):
         external_id = res['PersonID'][0]
 
         # Create user <-> external id link.
-        oauth_link_external_id(user, dict(id=external_id, method="cern"))
+        oauth_link_external_id(user, dict(id=external_id, method='cern'))
 
     groups = fetch_groups(res['Group'])
-    session['identity.cern_provides'] = [RoleNeed(group) for group in groups]
+    provides = [UserNeed(user.email)] + \
+               [RoleNeed(group + '@cern.ch') for group in groups]
```
egabancho: can you use `'{0}@cern.ch'.format(group)` instead? you know ... safety first ;-)

jirikuncar:
@omelkonian please use labels (FIX, NOTE) only when the message is important for readers of release notes. In the case of first commit you can remove the FIX label as it is just styling (`* FIX Converts strings to be more uniform ...`) and if you would really like to announce such change then label BETTER [1] is more appropriate.

--
[1] http://invenio.readthedocs.io/en/latest/technology/git.html#r2-remarks-on-commit-log-messages

egabancho:
Besides @jirikuncar's comment LGTM

jirikuncar:
@omelkonian please keep the branch up-to-date with the master branch and have a look to unresolved comments.

PR #75, opened 2016-07-21
global: sqlalchemy relationships refactor

krzysztof:
* Fixed the backref relationships in models.

Signed-off-by: Krzysztof Nowak <k.nowak@cern.ch>

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2014, 2015 CERN.
+# Copyright (C) 2016 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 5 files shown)


lnielsen:
Which problems does this fix solve?

krzysztof:
None, just convenience of having consistent ORM definitions across modules.

tests/__init__.py
```diff
@@ -16,3 +16,4 @@
 # You should have received a copy of the GNU General Public License
 # along with Invenio; if not, write to the Free Software Foundation, Inc.,
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+"""OAuthClient unit tests."""
```
jirikuncar: We should remove `__init__.py` file from `tests`.

jirikuncar:
@krzysztof `s/refactor/refactoring/`, `s/Fixed/Fixes/`

tests/__init__.py
krzysztof: Actually it's needed for importing ``tests/helpers.py``: https://github.com/inveniosoftware/invenio-oauthclient/blob/master/tests/test_contrib_cern.py#L32

jirikuncar: `from helper import ...` will work without `__init__.py`

PR #83, opened 2016-08-05
contrib: disconnect handler to unlink CERN account

### LINKED ISSUE #82: cern: missing disconnect handler to unlink account
This error occurs when logging in via cern. 
It is caused when this method is called: 
https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/contrib/cern.py#L236

To reproduce it Sign in, Sign out and then again sign in and the error should appear.

jirikuncar:
* Adds missing disconnect handler which unlinks CERN account.
  (closes #82)

Signed-off-by: Jiri Kuncar <jiri.kuncar@cern.ch>

--
cc @ioannistsanaktsidis

### DIFF
--- a/invenio_oauthclient/contrib/cern.py
+++ b/invenio_oauthclient/contrib/cern.py
@@ -86,11 +86,13 @@ In templates you can add a sign in/up link:
 import copy
 import re
 
-from flask import current_app, session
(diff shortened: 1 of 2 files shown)


egabancho:
LGTM

PR #88, opened 2016-08-19
contrib: CERN account info fix

pamfilos:
* ~~a~~A ❗  dds to session 'identity.provides','identity.cern_provides'**.**❗ 

Signed-off-by: Pamfilos Fokianos <pamfilosf@gmail.com>

### DIFF
--- a/invenio_oauthclient/contrib/cern.py
+++ b/invenio_oauthclient/contrib/cern.py
@@ -209,11 +209,18 @@ def get_resource(remote):
 def account_info(remote, resp):
     """Retrieve remote account information used to find local user."""
     res = get_resource(remote)
+    session.pop('cern_resource')
 
     email = res['EmailAddress'][0]
     external_id = res['PersonID'][0]
     nice = res['CommonName'][0]
     name = res['DisplayName'][0]
+
+    groups = fetch_groups(res['Group'])
+    provides = [UserNeed(email)] + \
+               [RoleNeed('{0}@cern.ch'.format(group)) for group in groups]
+
+    session['identity.cern_provides'] = provides
 
     return dict(
         user=dict(
@@ -257,11 +264,6 @@ def account_setup(remote, token, resp):
 
         # Create user <-> external id link.
         oauth_link_external_id(user, dict(id=external_id, method='cern'))
-
-    groups = fetch_groups(res['Group'])
-    provides = [UserNeed(user.email)] + \
-               [RoleNeed('{0}@cern.ch'.format(group)) for group in groups]
-    session['identity.cern_provides'] = provides
 
 
 @identity_loaded.connect
--- a/invenio_oauthclient/utils.py
+++ b/invenio_oauthclient/utils.py
@@ -73,8 +73,8 @@ def oauth_get_user(client_id, account_info=None, access_token=None):
                 id=external_id['id'], method=external_id['method']).first()
             if user_identity:
                 return user_identity.user
-        if account_info.get('email'):
-            return User.query.filter_by(email=account_info['email']).first()
+        if account_info.get('user').get('email'):
+            return User.query.filter_by(email=account_info['user']['email']).first()
     return None
 
 


invenio_oauthclient/utils.py
```diff
@@ -117,7 +117,8 @@ def oauth_link_external_id(user, external_id=None):
                 id_user=user.id
             ))
     except IntegrityError:
-        raise AlreadyLinkedError(user, external_id)
+        if external_id['method'] is not 'cern':
```
lnielsen: Please try to fix the problem which is [here](https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/utils.py#L76-L77):

```python
if account_info.get('user', {}).get('email'):
    return User.query.filter_by(email=account_info['user']['email']).first()
```

invenio_oauthclient/contrib/cern.py
```diff
@@ -209,12 +209,19 @@ def get_resource(remote):
...
     """Retrieve remote account information used to find local user."""
     res = get_resource(remote)
+    session.pop('cern_resource')
 
     email = res['EmailAddress'][0]
     external_id = res['PersonID'][0]
     nice = res['CommonName'][0]
     name = res['DisplayName'][0]
 
+    groups = fetch_groups(res['Group'])
```
lnielsen: Ping @egabancho  - please check this.

invenio_oauthclient/contrib/cern.py
```diff
@@ -209,12 +209,19 @@ def get_resource(remote):
...
+    provides = [UserNeed(email)] + \
+               [RoleNeed('{0}@cern.ch'.format(group)) for group in groups]
+
+    session['identity.cern_provides'] = provides
```
lnielsen: Not sure if this is safe to put it to the session - the user has not been authenticated at this point in time.

lnielsen: I think one possible attack could be something along the lines of:

1) Try to login as Alice who is member of CMS.
2) Now, login with Bob which is not member of CMS via different authentication mechanism (either local login or another oauth). -> Now Bob is possibly a member of CMS.

Essentially the trick is to get the groups stored on the session, and get the ``identity_loaded`` signal fired in some different manner.

Why are you in need of the groups at this point in time? Perhaps there's a different way.

invenio_oauthclient/contrib/cern.py
jirikuncar: IMHO it should not happen here but in `account_setup` as it was done before.

pamfilos: but if it happens in ```account_setup``` it will only be added to the sessions when you register.. if you go and log in from a different browser it won't be added

invenio_oauthclient/contrib/cern.py
pamfilos: It is the same way it was happening to ```account_setup``` and it passes the groups to the session..then on ```identity_loaded``` it updates the ```identity.provides``` of flask-principal, as before

We need to provide special permissions to users that belong to one e-group... Any other way is welcome

lnielsen: The difference here is that ``account_setup`` is called *after* a user has been authenticated, whereas ``account_info`` is called **before** a user has been authenticated. Hence, by doing it in ``account_setup`` you don't stick the list of groups to the session unless you are sure the user actually belongs to those groups.

invenio_oauthclient/contrib/cern.py
lnielsen: Ok, I understand what the problem is now. The solution is not to move the code to ``account_info`` because it opens up the code for attacks.

I'm not sure what would be best way to solve it, but it could be to do the groups fetching in a [Flask-Login user_logged_in signal](https://flask-login.readthedocs.io/en/latest/#signals)

CDS (@egabancho @ludmilamarian) I think must have the same problem or already have it solved.

invenio_oauthclient/utils.py
```diff
@@ -73,8 +73,8 @@ def oauth_get_user(client_id, account_info=None, access_token=None):
                 id=external_id['id'], method=external_id['method']).first()
             if user_identity:
                 return user_identity.user
-        if account_info.get('email'):
-            return User.query.filter_by(email=account_info['email']).first()
+        if account_info.get('user').get('email'):
```
jirikuncar: This is a good catch! Can you create a test-case for it?

jirikuncar:
```python
email  = account_info.get('user', {}).get('email')
if email:
    return User.query.filter_by(email=email).one_or_none()
```

invenio_oauthclient/contrib/cern.py
ioannistsanaktsidis: @lnielsen just tried adding the groups by using user_logged_in signal. Seems that it solves this issue and we do not have to touch the account_info code.

pamfilos:
closing since superseded by #89

PR #96, opened 2016-09-02
contrib: automatic CERN groups updates

jirikuncar:
Signed-off-by: Jiri Kuncar <jiri.kuncar@cern.ch>


invenio_oauthclient/contrib/cern.py
```diff
@@ -247,36 +274,39 @@ def disconnect_handler(remote, *args, **kwargs):
...
+        account_groups(token.remote_account, resource)
         user = token.remote_account.user
 
         # Create user <-> external id link.
         oauth_link_external_id(user, dict(id=external_id, method='cern'))
 
 
-@identity_loaded.connect
-def on_identity_loaded(sender, identity):
+@identity_changed.connect
```
ioannistsanaktsidis: Should we change this to ```@identity_changed``` ? By doing this the first time the user registers the groups are not available.

jirikuncar: @ioannistsanaktsidis yes, we don't want to call this method in each request.

ioannistsanaktsidis: Ok I see. The only problem is that in the register 
```
account = RemoteAccount.get(
        user_id=current_user.get_id(),
        client_id=client_id,
    )
```
returns ```None``` and the groups can't be fetched. Any ideas?

jirikuncar: @ioannistsanaktsidis please check latest updated version of this PR and code around https://github.com/inveniosoftware/invenio-oauthclient/pull/96/files#diff-b49dd27458efb60f2de74978e6d3eecfR304

ioannistsanaktsidis: @jirikuncar with the latest updates the issue is solved. LGTM!

ioannistsanaktsidis:
@jirikuncar  as  #97 is merged, is there a way to also include this PR or to create a new release?

jirikuncar:
@ioannistsanaktsidis this PR was ignored. I will rebase and do a new release straight away.

invenio_oauthclient/contrib/cern.py
tiborsimko: @egabancho Can you check the CDS LABS conditions?

lnielsen:
@jirikuncar We are on a extremely tight schedule right now and merging #97 was not to ignore this PR, just simply that we didn't have time for ping-pong. We have a long list of outstanding issues to do before Monday, and saving 10 minutes + waiting time here are important at this stage. Releases are cheap and easy to do so really don't see any problem in that this goes in a release couple of hours later.

ioannistsanaktsidis:
@jirikuncar  although the groups are added in the ```identity``` in your part, I just tested from the overlay side and ```identity.provides``` is empty.

jirikuncar:
@ioannistsanaktsidis can you check if you are not overriding the `identity` with any `identity_loaded` or `identity_changed` signal handler?

ioannistsanaktsidis:
@jirikuncar I think we should also add here https://github.com/jirikuncar/invenio-oauthclient/blob/840ec74a6553f4909db1c5cfda595a0da57ca781/invenio_oauthclient/contrib/cern.py#L279
 this 
```    session[OAUTHCLIENT_CERN_SESSION_KEY] = set([])
```
in case someone wants to just disconnect his account. In that case the groups should not be available in the session.

invenio_oauthclient/contrib/cern.py
```diff
@@ -185,9 +201,47 @@ def fetch_groups(groups):
...
+def extend_identity(identity, groups):
+    """Extend identity with roles based on CERN groups."""
+    provides = set([UserNeed(current_user.email)] + [
+        RoleNeed('{0}@cern.ch'.format(name)) for name in groups
+    ])
+    identity.provides |= provides
+    session[OAUTHCLIENT_CERN_SESSION_KEY] = provides
+
+
+def disconnect_identity(identity):
```
jirikuncar: @ioannistsanaktsidis what about something like this?

ioannistsanaktsidis: @jirikuncar it works good. I think we are done for now with this. Should we merge it  and create a new release?

jirikuncar: @ioannistsanaktsidis yes. I'm waiting for Travis CI and then I will update #99.

PR #107, opened 2016-10-20
contrib: save full name from orcid

Panos512:
- Stores the authors full name on orcid sign-up.

Signed-off-by: Panos Paparrigopoulos panos.paparrigopoulos@cern.ch

### DIFF
--- a/invenio_oauthclient/contrib/orcid.py
+++ b/invenio_oauthclient/contrib/orcid.py
@@ -156,7 +156,11 @@ def account_info(remote, resp):
     .. code-block:: python
 
         {
-            'user': {},
+            'user': {
+                'profile': {
+                    'full_name': 'Full Name',
+                },
+            },
             'external_id': 'github-unique-identifier',
             'external_method': 'github',
         }
@@ -167,11 +171,15 @@ def account_info(remote, resp):
     """
     orcid = resp.get('orcid')
 
-    return dict(
-        external_id=orcid,
-        external_method='orcid',
-        user=dict()
-    )
+    return {
+        'external_id': orcid,
+        'external_method': 'orcid',
+        'user': {
+            'profile': {
+                'full_name': resp.get('name'),
+            },
+        },
+    }
 
 
 def disconnect_handler(remote, *args, **kwargs):
@@ -187,7 +195,7 @@ def disconnect_handler(remote, *args, **kwargs):
     orcid = account.extra_data.get('orcid')
 
     if orcid:
-        oauth_unlink_external_id(dict(id=orcid, method='orcid'))
+        oauth_unlink_external_id({'id': orcid, 'method': 'orcid'})
     if account:
         with db.session.begin_nested():
             account.delete()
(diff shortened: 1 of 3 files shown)


invenio_oauthclient/contrib/orcid.py
```diff
@@ -205,9 +205,14 @@ def account_setup(remote, token, resp):
...
         # Retrieve ORCID from response.
         orcid = resp.get('orcid')
-
+        full_name = resp.get('name')
+        
         # Set ORCID in extra_data.
-        token.remote_account.extra_data = {'orcid': orcid}
+        token.remote_account.extra_data = {
+            'orcid': orcid,
+            'full_name': full_name
```
Panos512: For my case I just need access to the `authors full_name` from the `current_user` flask variable. 

The name could also be stored on the user profile (and it might should be added there automatically on `oauth-signup`). 

The only reason I see, right now, for storing the `orcid-full_name` separately on the `extra_data` is that since the user can actually edit his full_name on his invenio-userprofile (after wedding e.g) we might end up with two different names for the same person.

jirikuncar: So shall we extend the `user=dict()` with `user={'profile': {'full_name': resp.get('name')}` on line 173?

Panos512: done! 💃

invenio_oauthclient/contrib/orcid.py
```diff
@@ -170,7 +170,11 @@ def account_info(remote, resp):
     return dict(
         external_id=orcid,
         external_method='orcid',
-        user=dict()
+        user={
+            'profile': {
+                'full_name': resp.get('name')
```
jirikuncar: Can you add commas at the end of line after which new item can be added?

(175, 176, 177)

--
Can we use `{}` instead of `dict()` also at 170? I see that it's used inconsistently in the file (not your fault, but we should improve it when touching the file). Thanks!

invenio_oauthclient/contrib/orcid.py
```diff
@@ -167,11 +167,15 @@ def account_info(remote, resp):
     """
     orcid = resp.get('orcid')
 
-    return dict(
+    return {
         external_id=orcid,
```
jirikuncar: syntax issue?

tests/test_contrib_orcid.py
```diff
@@ -128,6 +132,7 @@ def test_authorized_signup(userprofiles_app, example_orcid, orcid_bio):
             id=example_data['orcid']
         ).one()
         # FIXME see contrib/orcid.py line 167
+        assert user.profile.full_name == 'Josiah Carberry'
```
Panos512: @jirikuncar  If I understood correctly that's the place for the check. The problem is the `full_name` is empty. I am a bit lost in the packages, but if I get it right `invenio-userprofiles` should use the `account_info` function to populate the profile (resulting to 'Josiah Carberry' filling the `full_name` field).

jirikuncar: Can you check if we are using Invenio-UserProfile in tests?

Panos512: Yes, we initialise the `userprofiles_app` in `conftest.py` and adding the dependency on `setup.py:tests_require`

jirikuncar: Can you check it locally with https://github.com/inveniosoftware/invenio-userprofiles/pull/65 ?

Panos512: I did and it didn't fix it. 😞

invenio_oauthclient/contrib/orcid.py
```diff
@@ -156,7 +156,11 @@ def account_info(remote, resp):
     .. code-block:: python
 
         {
-            'user': {},
+            'user': {
+                'profile': {
+                    'full_name': 'user's-full-name',
```
jirikuncar: please escape the inner quote `'` or change the text to e.g. 'Full Name'.

jirikuncar:
LGTM

PR #108, opened 2016-11-02
global: removal of Flask-Security usage

hachreak:
* Removes Flask-Security usage with other alternatives from already
  existing dependencies.

Signed-off-by: Leonardo Rossi <leonardo.r@cern.ch>

### DIFF
--- a/requirements-devel.txt
+++ b/requirements-devel.txt
@@ -22,6 +22,7 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
+-e git+git://github.com/inveniosoftware/flask-security-fork.git#egg=flask-security-fork
 -e git+git://github.com/inveniosoftware/invenio-accounts.git#egg=invenio-accounts
 -e git+git://github.com/inveniosoftware/invenio-db.git#egg=invenio-db
 # -e git+git://github.com/inveniosoftware/invenio-upgrader.git#egg=invenio-upgrader
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ tests_require = [
 
 extras_require = {
     'admin': [
-        'invenio-admin>=1.0.0a2',
+        'invenio-admin>=1.0.0a3',
     ],
     'docs': [
         'Sphinx>=1.4.2',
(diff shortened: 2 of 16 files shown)


setup.py
```diff
@@ -85,8 +85,9 @@
 install_requires = [
     'Flask-BabelEx>=0.9.3',
     'Flask-Breadcrumbs>=0.3.0',
+    'Flask-Login>=0.4.0',
     'Flask-OAuthlib>=0.9.3',
-    'Flask-Security>=1.7.5',
+    'Flask-Security-Fork>=1.8.0',
```
lnielsen: Do we still need the dependency when you are removing all use of it?

hachreak: I removed many, but not all. T_T
```
16:38 $ git grep flask_security
invenio_oauthclient/utils.py:from flask_security.confirmable import requires_confirmation
invenio_oauthclient/utils.py:from flask_security.registerable import register_user
tests/test_handlers.py:from flask_security.confirmable import _security
tests/test_utils.py:from flask_security.confirmable import _security
```

I still have one test failing, but no idea why.
It looks like that after `login_user()` the `g.identity` is still `AnonymousIdentity`. 

Any idea? :)

jirikuncar: We can not remove the `flask_security` here since we rely on its behavior.

hachreak:
@jirikuncar  I reverted the changes for `login_user` and `logout_user`.
Now tests are passing.

setup.py
```diff
@@ -85,14 +85,15 @@
 install_requires = [
     'Flask-BabelEx>=0.9.3',
     'Flask-Breadcrumbs>=0.3.0',
+    'Flask-Login>=0.4.0',
```
jirikuncar: can you keep the import pinned to same number as in Invenio-Accounts and do:

```python
try:
    from flask_login.utils import _create_identifier
except ImportError:
    from flask_login import _create_identifier
```
instead. Thanks!

jirikuncar: Can you remove the monkey patching and provide deterministic request context with remote addr and User-Agent?

hachreak: Deterministic request? What do you mean?
Anyway, could we open a ticket for it? Thanks.
I'm quite busy now with the CDS sprint and this PR with this [PR for deposit](https://github.com/inveniosoftware/invenio-deposit/pull/143) are necessary to us to follow working.
When I'll be free I'll better check.

requirements-devel.txt
```diff
@@ -22,6 +22,7 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
+-e git+git://github.com/inveniosoftware/flask-security-fork.git#egg=flask-security
```
hachreak: done

hachreak:
ping @jirikuncar

egabancho:
LGTM

setup.py
jirikuncar: What about adding

```python
try:
    from flask_login.utils import _create_identifier
except ImportError:
    from flask_login import _create_identifier
```

to `_compat.py` and keep `Flask-Login>=0.3.5`?

hachreak:
ping @jirikuncar I set `>0.3.2`. :shipit: ?

hachreak:
@jirikuncar :shipit: ?

jirikuncar:
LGTM

PR #119, opened 2017-02-09
Alembic mysql fix

nikofil:


drjova:
LGTM

invenio_oauthclient/alembic/97bbc733896c_create_oauthclient_tables.py
```diff
@@ -73,6 +73,7 @@ def upgrade():
         sa.Column('secret', sa.Text(), nullable=False),
         sa.ForeignKeyConstraint(
             ['id_remote_account'], [u'oauthclient_remoteaccount.id'],
+            name='fk_oauthclient_remote_token_remote_account'
```
lnielsen: I agree that this should normally be the case, however I think since no production installations actually uses alembic upgrades yet it is not worth the effort.

nikofil: The issue here is that on MySQL databases, which support constraint names of up to 60 characters, this alembic recipe will fail anyway (as the name is longer), so the migration will fail before getting to the other recipe.

jirikuncar: This is easy case where one can learn how to create simple Alembic recipe and present this PR as example how things should be done.

lnielsen: I as understand the upgrade is actually broken as it is today since it will never actually create the table on e.g. MySQL because the naming conventions where introduced after this upgrade was merged? I.e. there won't be a table on MySQL to upgrade.

jirikuncar: That's just for MySQL. It works on Postgres, so you want to make sure that you change the name for Postgres users too. Otherwise it might backfire in the future...

egabancho: I understand @jirikuncar's point and if the package were "fully" released I would agreed with him, but given the fact that the package is in `alpha` stage I would just update the recipe as @nikofil did.

PR #126, opened 2017-04-06
entry point function for SECRET_KEY change

dinosk:
Signed-off-by: Dinos Kousidis <konstantinos.kousidis@cern.ch>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ tests_require = [
     'pytest-cache>=1.0',
     'pytest-cov>=1.8.0',
     'pytest-pep8>=1.0.6',
-    'pytest>=2.8.0',
+    'pytest>=2.8.3',
     'simplejson>=3.8',
 ]
 
@@ -60,14 +60,14 @@ extras_require = {
         'uritemplate.py>=0.2.0,<2.0',
     ],
     'mysql': [
-        'invenio-db[mysql]>=1.0.0b3',
+        'invenio-db[mysql]>=1.0.0b5',
     ],
     'orcid': [],
     'postgresql': [
-        'invenio-db[postgresql]>=1.0.0b3',
+        'invenio-db[postgresql]>=1.0.0b5',
     ],
     'sqlite': [
-        'invenio-db>=1.0.0b3',
+        'invenio-db>=1.0.0b5',
     ],
     'tests': tests_require,
 }
@@ -93,7 +93,7 @@ install_requires = [
     'blinker>=1.4',
     # fix issue with cryptography old version and cffi new version
     'cryptography>=1.5',  # sqlalchemy-utils dependency
-    'invenio-accounts>=1.0.0b1',
+    'invenio-accounts>=1.0.0b3',
     'invenio-mail>=1.0.0a5',
     'six>=1.9',
     'sqlalchemy-utils>=0.31',
@@ -142,6 +142,10 @@ setup(
         ],
         'invenio_db.models': [
             'invenio_oauthclient = invenio_oauthclient.models',
+        ],
+        'invenio_base.secret_key': [
+            'invenio_oauthclient = '
+            'invenio_oauthclient.utils:rebuild_access_tokens',
         ],
         'invenio_i18n.translations': [
             'invenio_oauthclient = invenio_oauthclient',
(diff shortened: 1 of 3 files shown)


invenio_oauthclient/receivers.py
```diff
@@ -0,0 +1,57 @@
...
+@secret_key_changed.connect
+def notify_secret_key_changed(sender, old_key):
+    """Rebuild the access_token field when the SECRET_KEY is changed.
+    
+    Needed to fix the access tokens used in the login.
+
+    :param sender: signal sender object.
+    :param old_key: the old SECRET_KEY.
+    """
+    print('invenio-oauthclient: changing SECRET_KEY...')
```
jirikuncar: Since `sender` should be `app`, you can use `(app|sender).logger.info`.

invenio_oauthclient/receivers.py
```diff
@@ -0,0 +1,57 @@
...
+    db.session.close_all()
+    eng = db.get_engine(current_app)
+    conn = eng.connect()
+    query_result = conn.execute("""select id_remote_account, access_token 
+                                   from oauthclient_remotetoken;""")
```
jirikuncar: Please avoid using text based queries and use the model instead.

dinosk: That is what I tried at first but it results in a decoding error, as the model will try to decrypt the access_token using the new SECRET_KEY from app.config.

invenio_oauthclient/receivers.py
dinosk: @jirikuncar I see, thank you.

invenio_oauthclient/receivers.py
```diff
@@ -0,0 +1,39 @@
...
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+from invenio_base.signals import secret_key_changed
+from invenio_oauthclient.models import RemoteToken
+from invenio_base.utils import rebuild_encrypted_field
+
+
+@secret_key_changed.connect
```
nharraud: We had issues in the past with side effects of importing files. As this code is optional it would be better to call explicitly `secret_key_changed.connect(secret_key_changed.connect)` instead of using it as a decorator. Also this file should be in a `contrib` directory as it depends on `invenio_base`.

jirikuncar: Please consider using entry points over signals so it is easier to find out which packages are extending the CLI functionality.

tests/test_contrib_utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # Changing application SECRET_KEY
+    app.secret_key = "NEW_SECRET_KEY"
+
+    # Asserting the decoding error occurs with the stale SECRET_KEY
+    with pytest.raises(UnicodeDecodeError):
+        RemoteToken.query.all()
+
+    rebuild_access_tokens(app, old_secret_key)
+    remote_token = RemoteToken.query.all()[0]
```
nharraud: The returned first RemoteToken is not necessarily the one you expect. It would be better to query for it specifically.

invenio_oauthclient/contrib/utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+def rebuild_access_tokens(caller, old_key):
+    """Rebuild the access token field when the SECRET_KEY is changed.
+
+    Fixes users' login
+
+    :param sender: signal sender object.
+    :param old_key: the old SECRET_KEY.
+    """
+    caller.logger.info('rebuilding RemoteToken.access_token...')
```
nharraud: I think you can use current_app here. This code is called in the application context. As said in inveniosoftware/invenio-base#107 we can remove the `caller` parameter as we are not using signals anymore.

invenio_oauthclient/contrib/utils.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
```
jirikuncar: IMHO we can move it from the contrib folder now.

setup.py
```diff
@@ -44,7 +44,7 @@
     'pytest-cache>=1.0',
     'pytest-cov>=1.8.0',
     'pytest-pep8>=1.0.6',
-    'pytest>=2.8.0',
+    'pytest>=2.8.3',
```
dinosk: Bumped because I ran into this: https://github.com/pytest-dev/pytest/issues/995

lnielsen: Will you perhaps add an entry to inveniosoftware/troubleshooting?

PR #127, opened 2017-04-11
installation: api_apps application registering

egabancho:
Signed-off-by: Leonardo Rossi <leonardo.r@cern.ch>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -132,6 +132,9 @@ setup(
         'invenio_base.apps': [
             'invenio_oauthclient = invenio_oauthclient:InvenioOAuthClient',
         ],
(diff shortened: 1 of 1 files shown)


nharraud:
Why do we need to register the application on the API app again?
In B2SHARE we do it ourselves in out setup.py [here](https://github.com/EUDAT-B2SHARE/b2share/blob/d69eabbca6291783f04da049d8412ef12e49c5a1/setup.py#L192) and [here](https://github.com/EUDAT-B2SHARE/b2share/blob/d69eabbca6291783f04da049d8412ef12e49c5a1/setup.py#L186).
Do we want this to be the default behavior?

PR #132, opened 2017-05-18
cern: replaces 'PersonID' to 'uidNumber'

### LINKED ISSUE #112: cern: PersonID is not an unique identifier
To recognize CERN user, field PersonID is used (as an external id). This field is the same for all accounts associated with the same user (primary, secondary and service accounts). 
Field uidNumber can be used instead.

pamfilos:
* closes #112

Signed-off-by: Pamfilos Fokianos <pamfilosf@gmail.com>

### DIFF
--- a/invenio_oauthclient/contrib/cern.py
+++ b/invenio_oauthclient/contrib/cern.py
@@ -270,7 +270,7 @@ def account_info(remote, resp):
     resource = get_resource(remote)
 
     email = resource['EmailAddress'][0]
-    external_id = resource['PersonID'][0]
+    external_id = resource['uidNumber'][0]
     nice = resource['CommonName'][0]
     name = resource['DisplayName'][0]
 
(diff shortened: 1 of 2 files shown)


tiborsimko:
@egabancho Can you check whether this change would work for CDS? Or should we make it configurable for each service?

lnielsen:
- [ ] Replace field with config variable that can be set to either PersonID or uidNumber
- [ ] Document the difference between the two attributes.

egabancho:
> Replace field with config variable that can be set to either PersonID or uidNumber

Do we really need this? Using `PersonID` might leed to accounts having wrong access rights, among other small problems.

lnielsen:
We just had a discussion about this in the Sprint kick-off. As I understood on @drjova you needed to use ``PersonID`` on CDS (as it represents the same user over multiple accounts), and CAP needs ``uidNumber``. 

If that's not the case, then I think we can just merge?

egabancho:
This was de initial idea, but it might cause us a few headaches, so, `uidNumber` for everyone!

Once tests are passing we can :shipit:

PR #140, opened 2017-07-31
admin: disable linked accounts menu item

hjhsalo:
* do not show Linked Accounts -menuitem if no OAuth providers
  have been configured.

Signed-off-by: Harri Hirvonsalo <harri.hirvonsalo@cern.ch>

### DIFF
--- a/invenio_oauthclient/views/settings.py
+++ b/invenio_oauthclient/views/settings.py
@@ -48,7 +48,9 @@ blueprint = Blueprint(
     _('%(icon)s Linked accounts', icon='<i class="fa fa-link fa-fw"></i>'),
     order=3,
     active_when=lambda: request.endpoint.startswith(
-        'invenio_oauthclient_settings.')
+        'invenio_oauthclient_settings.'),
+    visible_when=lambda: bool(current_app.config.get(
+        'OAUTHCLIENT_REMOTE_APPS')) is not False
 )
 @register_breadcrumb(
     blueprint, 'breadcrumbs.settings.oauthclient', _('Linked accounts')
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,8 @@ services:
 
 language: python
 
+dist: precise
+
 matrix:
   fast_finish: true
 


invenio_oauthclient/views/settings.py
```diff
@@ -41,14 +41,23 @@
...
 @blueprint.route('/', methods=['GET', 'POST'])
 @login_required
 @register_menu(
     blueprint, 'settings.oauthclient',
     _('%(icon)s Linked accounts', icon='<i class="fa fa-link fa-fw"></i>'),
     order=3,
     active_when=lambda: request.endpoint.startswith(
-        'invenio_oauthclient_settings.')
+        'invenio_oauthclient_settings.'),
+    visible_when=menu_item_visible
```
drjova: maybe we can use something like the following?  WDYT :)

```python
lambda: bool(current_app.config.get('OAUTHCLIENT_REMOTE_APPS')) is not False
```

hjhsalo: You are right, this is more in line to the previously used coding style. Changed to this one. Thanks.

drjova:
@hjhsalo thanks! Could you please also rebase?

drjova:
@hjhsalo thanks a lot! One more small thingy could you please keep the commit subject line to [50 chars](https://invenio.readthedocs.io/en/latest/community/contribution-guide.html#commit-messages)? :P

invenio_oauthclient/version.py
```diff
@@ -28,4 +28,4 @@
 and parsed by ``setup.py``.
 """
 
-__version__ = '1.0.0a14.dev20170511'
+__version__ = '1.0.0a14.dev20170802'
```
drjova: we don't need to change the version here :)

hjhsalo: OK? Could you elaborate a bit?

hjhsalo: > One more small thingy could you please keep the commit subject line to 50 chars? :P

I'm all out of ideas for clean, short and unambiguous title. Do you have suggestions?

hjhsalo:
> we don't need to change the version here :)

OK? Could you elaborate a bit?

drjova:
> we don't need to change the version here :)
>> OK? Could you elaborate a bit?

We change the version only when we make a new release, for [example](https://github.com/inveniosoftware/invenio-oauthclient/commit/3b9b9df91b7b1a2529f527ead37ab49d1b1717e9). So there is no need to update the version now :)

hjhsalo:
> We change the version only when we make a new release, for example. So there is no need to update the version now :)

OK. Thanks

PR #148, opened 2017-08-08
handlers: fill csrf token on auto signup form

### LINKED ISSUE #115: cern: `account_setup` not working
For some reason [this](https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/contrib/cern.py#L319)
```python
assert not isinstance(g.identity, AnonymousIdentity)
```
evaluates to `False`. (To investigate more in depth)

cc @pamfilos

diegodelemos:
* Auto fill csrf_token to avoid csrf validation error on self
  submission. csrf_token validation flag differs from Flask-WTF
  version 0.13 to 0.14. Moreover, the old flag will be removed on
  Flask-WTF 1. (closes #115)

Signed-off-by: Diego Rodriguez <diego.rodriguez@cern.ch>

### DIFF
--- a/invenio_oauthclient/handlers.py
+++ b/invenio_oauthclient/handlers.py
@@ -42,8 +42,9 @@ from .models import RemoteAccount, RemoteToken
 from .proxies import current_oauthclient
 from .signals import account_info_received, account_setup_committed, \
     account_setup_received
-from .utils import disable_csrf, fill_form, oauth_authenticate, \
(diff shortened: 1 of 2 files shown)


diegodelemos:
@hachreak autofilling the `csrf` token was breaking on `flask-wtf` > 0.14 so I had to add a function to decide which flag needs to be passed in order to skip the `csrf` validation

PR #150, opened 2017-08-09
version: post-release bump

drjova:
after https://github.com/inveniosoftware/invenio-oauthclient/pull/149


invenio_oauthclient/version.py
```diff
@@ -28,4 +28,4 @@
 and parsed by ``setup.py``.
 """
 
-__version__ = '1.0.0a14.dev20170511'
+__version__ = '1.0.0b2.dev20170809'
```
nharraud: The post release bump starts with the version number which will be released next. In this case I doubt that we will skip Beta 1.

PR #157, opened 2018-01-22
utils: Change redirect uri check

zzacharo:
Returns always relative paths except if the `APP_ALLOWED_HOSTS`
variable is set, thus checks the host to be included in that list.

### DIFF
--- a/invenio_oauthclient/utils.py
+++ b/invenio_oauthclient/utils.py
@@ -160,18 +160,6 @@ def oauth_unlink_external_id(external_id):
                                      method=external_id['method']).delete()
 
 
-def is_local_url(target):
-    """Determine if URL is a local.
-
-    :param target: The URL to check.
-    :returns: ``True`` if the target is a local url.
-    """
-    server_name = current_app.config['SERVER_NAME']
-    test_url = urisplit(target)
-    return not test_url.host or test_url.scheme in ('http', 'https') and \
-        server_name == test_url.host
-
-
 def get_safe_redirect_target(arg='next'):
     """Get URL to redirect to and ensure that it is local.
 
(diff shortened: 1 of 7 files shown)


zzacharo:
closes CERNDocumentServer/cds-videos#1488

invenio_oauthclient/utils.py
```diff
@@ -160,27 +160,20 @@ def oauth_unlink_external_id(external_id):
...
 
     :param arg: URL argument.
     :returns: The redirect target or ``None``.
     """
     for target in request.args.get(arg), request.referrer:
-        if target and is_local_url(target):
-            return target
+        if target:
+            target = urisplit(target)
+            if target.path:
```
slint: Could this be an issue for redirects to the root path (ie. for `https://videos.cern.ch`, `target.path == ''`)?

invenio_oauthclient/utils.py
```diff
@@ -160,27 +160,20 @@ def oauth_unlink_external_id(external_id):
...
+                allowed_hosts = current_app.config.get('APP_ALLOWED_HOSTS', [])
+                if target.host in allowed_hosts:
```
slint: So, this is the safe and dumb-proof way, but might come in the way of people from testing/debugging things... @lnielsen @egabancho WDYT?

invenio_oauthclient/utils.py
zzacharo: yes this should be integrated.

invenio_oauthclient/utils.py
zzacharo: getting in their way in what sense?

PR #158, opened 2018-01-30
Added Globus as an identity provider

NickolausDS:
* Globus Login work for any provider Globus Auth supports
* Added tests, coverage of new code is 100%
* Added Sphinx docs

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,7 @@
 MIT License
 
 Copyright (C) 2015-2018 CERN.
+Copyright (C) 2018 University of Chicago.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in
--- a/docs/examplesapp.rst
+++ b/docs/examplesapp.rst
@@ -28,3 +28,10 @@ CERN
 .. include:: ../examples/cern_app.py
    :start-after: SPHINX-START
    :end-before: SPHINX-END
+
+Globus
+------
+
+.. include:: ../examples/globus_app.py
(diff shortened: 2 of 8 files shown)


lnielsen:
Thanks a lot for the contribution. We'll have to keep the contribution on hold for a bit because we're changing the license of Invenio from GPL to MIT. Unfortunately, discussion with our lawyers is taking longer than expected, and thus we cannot integrate code from non-CERN contributors until this has finished. I'm sorry about this.

NickolausDS:
Thanks for the update, I hope the license transition goes smoothly.

slint:
@NickolausDS we can merge this, though could you tell us what legal entity is the copyright holder for this code, so we can include it in the headers? Also, please check if the case is that since this code might have been written as part of an employment contract which grants your employer the copyright to all source code you write (keep in mind that this could also apply for code written outside of working hours), the copyright holder might be your employer.

lnielsen:
@rpwagner @NickolausDS - I've rebased the PR and updated the copyright headers. Last thing before I press the merge button, could you please confirm that you license your PR under MIT License (since the old headers said GPL - which was probably just due to copy/paste)?

NickolausDS:
@lnielsen Yes, confirming here to license this pull request under MIT. Thanks for switching that.

lnielsen:
Merged. Thanks a lot for the contribution, and apologies for taking so long to merge.

NickolausDS:
No need for an apology, I know licensing can get complex and take time to sort out all the details. 

Thanks for picking this back up, your help was much appreciated!

PR #159, opened 2018-02-19
import EncryptedType correctly from sqlalchemy_utils

Dinika:
sqlalchemy_utils recently [updated their `encrypted_type` module](https://github.com/kvesteri/sqlalchemy-utils/commit/7009afe4c8bb9bc14df5ffd205ba4aeaffbe2b75). `EncryptedType` now needs to be imported from `sqlalchemy_utils.types.encrypted.encrypted_type`.

Signed-off-by: Dinika Saxena <dinika.saxena@cern.ch>

### DIFF
--- a/invenio_oauthclient/alembic/97bbc733896c_create_oauthclient_tables.py
+++ b/invenio_oauthclient/alembic/97bbc733896c_create_oauthclient_tables.py
@@ -45,7 +45,7 @@ def upgrade():
         sa.Column('client_id', sa.String(length=255), nullable=False),
         sa.Column(
             'extra_data',
-            sqlalchemy_utils.types.json.JSONType(),
+            sqlalchemy_utils.JSONType(),
             nullable=False),
         sa.ForeignKeyConstraint(['user_id'], [u'accounts_user.id'], ),
         sa.PrimaryKeyConstraint('id'),
@@ -69,7 +69,7 @@ def upgrade():
         sa.Column('token_type', sa.String(length=40), nullable=False),
         sa.Column(
             'access_token',
-            sqlalchemy_utils.types.encrypted.EncryptedType(),
+            sqlalchemy_utils.EncryptedType(),
             nullable=False),
         sa.Column('secret', sa.Text(), nullable=False),
         sa.ForeignKeyConstraint(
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -33,6 +33,7 @@
 
 tests_require = [
     'SQLAlchemy-Continuum>=1.2.1',
+    'SQLAlchemy-Utils~=0.0,>=0.33.0.',
```
lnielsen: This is in ``test_require`` but I think it should be in ``install_requires`` instead (where the dependency is already). However, if you apply the above change then we don't need any changes to setup.py.

invenio_oauthclient/models.py
```diff
@@ -30,7 +30,7 @@
 from sqlalchemy.ext.mutable import MutableDict
 from sqlalchemy.orm import backref
 from sqlalchemy_utils import JSONType
-from sqlalchemy_utils.types.encrypted import EncryptedType
+from sqlalchemy_utils.types.encrypted.encrypted_type import EncryptedType
```
lnielsen:
```python
try:
    # SQLAlchemy-Utils <=0.32
    from sqlalchemy_utils.types.encrypted import EncryptedType
except ImportError:
    # SQLAlchemy-Utils >=0.33
    from sqlalchemy_utils.types.encrypted.encrypted_type import EncryptedType
```

In order to not force existing installations to upgrade SQLAlchemy-Utils immediately, I would go with above. ``EncryptedType`` is being used in other modules (e.g. OAuth2Server I think), hence if not all modules are fixed and released at the same time then it will cause conflicts

lnielsen: Actually, on second look into SQLAlchemy-Utils, following should work on both versions:

```
from sqlalchemy_utils import EncryptedType
```

invenio_oauthclient/alembic/97bbc733896c_create_oauthclient_tables.py
```diff
@@ -69,7 +69,7 @@ def upgrade():
         sa.Column('token_type', sa.String(length=40), nullable=False),
         sa.Column(
             'access_token',
-            sqlalchemy_utils.types.encrypted.EncryptedType(),
+            sqlalchemy_utils.types.encrypted.encrypted_type.EncryptedType(),
```
david-caro: This should use the top level `sqlalchemy_utils.EncryptedType` one.

PR #168, opened 2018-03-20
models: add 'sqlalchemy_utils.models.Timestamp' mixin

### LINKED ISSUE #160: models: missing "sqlalchemy_utils.models.Timestamp" mixin
[Example](https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/models.py#L35) on [invenio_accounts.models.SessionActivity](https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/models.py#L118)

ChiaraBi:
* alembic: create recipe to add 'created' and 'updated' columns

* close #160

### DIFF
--- a/invenio_oauthclient/alembic/bff1f190b9bd_add_timestamp_oauthclient.py
+++ b/invenio_oauthclient/alembic/bff1f190b9bd_add_timestamp_oauthclient.py
@@ -1,0 +1,71 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2016-2018 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


invenio_oauthclient/alembic/bff1f190b9bd_add_timestamp_oauthclient.py
```diff
@@ -0,0 +1,59 @@
...
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    # Add 'created' and 'updated' columns to RemoteAccount
+    op.add_column(
+        'oauthclient_remoteaccount',
+        sa.Column('created', sa.DateTime(), nullable=False))
```
lnielsen: What will be the default value for people upgrading?

Issue #178, opened 2018-10-26
cern: lightweight account throws 500


ntarocco:
[This change](https://github.com/inveniosoftware/invenio-oauthclient/pull/132) has introduced a bug for lightweight accounts. CERN lightweight accounts do not have have a `uidNumber`, so when logging in you get a stack trace on `external_id`:

`KeyError: 'uidNumber'`
```
invenio_oauthclient/contrib/cern.py in account_info at line 262
def account_info(remote, resp):
    """Retrieve remote account information used to find local user."""
    resource = get_resource(remote)
    email = resource['EmailAddress'][0]
    external_id = resource['uidNumber'][0]
```

<details><summary>Click to see an example of a login with a lightweight CERN account</summary>
<p>
```
[
  {
    "Type": "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier",
    "Value": "<email>@gmail.com"
  },
  {
    "Type": "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/upn",
    "Value": "<email>@gmail.com"
  },
  {
    "Type": "http://schemas.xmlsoap.org/claims/UPN",
    "Value": "<email>@gmail.com"
  },
  {
    "Type": "http://schemas.xmlsoap.org/claims/EmailAddress",
    "Value": "<email>@gmail.com"
  },
  {
    "Type": "http://schemas.xmlsoap.org/claims/CommonName",
    "Value": "j34hwtjhtg54g64j646g4"
  },
  {
    "Type": "http://schemas.microsoft.com/ws/2008/06/identity/claims/role",
    "Value": "CERN Externals"
  },
  {
    "Type": "http://schemas.xmlsoap.org/claims/DisplayName",
    "Value": "<email>@gmail.com"
  },
  {
    "Type": "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name",
    "Value": "<email>@gmail.com"
  },
  { "Type": "http://schemas.xmlsoap.org/claims/PersonID", "Value": "-3243534" },
  {
    "Type": "http://schemas.xmlsoap.org/claims/IdentityClass",
    "Value": "Unverified External"
  },
  { "Type": "http://schemas.xmlsoap.org/claims/Federation", "Value": "CERN" },
  { "Type": "http://schemas.xmlsoap.org/claims/AuthLevel", "Value": "Normal" },
  {
[...]

PR #179, opened 2018-11-27
contrib: add extra cern data

zzacharo:


invenio_oauthclient/contrib/cern.py
```diff
@@ -258,8 +258,11 @@ def account_info(remote, resp):
     """Retrieve remote account information used to find local user."""
     resource = get_resource(remote)
 
+    person_id = resource.get('PersonID', [None])
+    external_id = resource.get('uidNumber', person_id)
+
     email = resource['EmailAddress'][0]
-    external_id = resource['uidNumber'][0]
+    external_id = external_id[0]
```
ntarocco: simply
```
resource.get('uidNumber', person_id[0])
```
?

lnielsen:
Ping @tiborsimko @pamfilos - Please check this PR as you are using the CERN OAuth contrib.

tests/test_contrib_cern.py
```diff
@@ -37,6 +38,22 @@ def test_fetch_groups(app, example_cern):
                for group in ('Group{}'.format(i) for i in range(4, 6)))
 
 
+def test_fetch_extra_data(app, example_cern):
```
ntarocco: Can you please add a test where the response does NOT contain some of the values too?
So we ensure that future changes will not break behaviour. Thanks!

invenio_oauthclient/contrib/cern.py
```diff
@@ -206,9 +220,12 @@ def account_groups(account, resource, refresh_timedelta=None):
         return account.extra_data.get('groups', [])
 
     groups = fetch_groups(resource['Group'])
+    extra_data = fetch_extra_data(resource)
```
zzacharo: Also @lnielsen @pamfilos @tiborsimko do you find any case that this should be configurable in order in the future to parse any field in the cern oauth response?

tests/test_contrib_cern.py
zzacharo: @ntarocco [here](https://github.com/inveniosoftware/invenio-oauthclient/pull/179/files#diff-25fc83081804f5e108a7dd26baf648beR57)

ntarocco:
I think this is closing https://github.com/inveniosoftware/invenio-oauthclient/issues/178 if you want to mention it

zzacharo:
@ntarocco no actually. I reverted the fix and I am gonna create a separate pr for that.

zzacharo:
Ping @lnielsen @tiborsimko @pamfilos could you please check it as I need this?

invenio_oauthclient/contrib/cern.py
pamfilos: @zzacharo I think since we are adding this feature, it's better to have it configurable

egabancho:
@zzacharo do you think we should write an upgrade recipe or something to update/invalidate all the accounts which don’t have this data?  Perhaps simply force to reload it using the last updated field.

zzacharo:
@egabancho yes that would be good I think. @lnielsen I don't think it will be incompatible especially if we write a recipe to update user's metadata with what @egabancho proposed..I don't know if I miss something...

ntarocco:
@egabancho the refresh time is a config of 5 minutes by default (if I am not wrong), I don't think any of us changed it? At least nor videos or cern analysis preservation...
@lnielsen I also don't think it is breaking since it is adding new fields and not changing existings...

Issue #180, opened 2018-11-29
cern: disconnection needs an extra redirect


ntarocco:
At CERN, the login process using OAuth creates a cookie when on the domain login.cern.ch.
This means that to complete a logout, that cookie needs to be removed.

Invenio-oauthclient cern is missing a redirect to `https://login.cern.ch/adfs/ls/?wa=wsignout1.0` when logging out so that cookie is deleted. Currently, we redirect to the homepage: https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/contrib/cern.py#L293

Without this extra step, the cookie is not cleaned.

ppanero:
Problem state:
------------------
- The logout workflow does not pass by the ``disconnect_handler``.
- Flask Security uses Flask Login to logout user. Flask Login [sends a signal](https://github.com/maxcountryman/flask-login/blob/3e521a326696cafbfbebfbb80a2fbffed68e6cf3/flask_login/utils.py#L215) to logout. However, that signal is in the middle of the process and therefore no redirection is possible.

Flask Security:
------------------
- Flask Security provides a configurable logout URL via ``SECURITY_POST_LOGOUT_VIEW``.
- This URL [must have the same domain](https://github.com/mattupstate/flask-security/blob/3e15d06ee82a728d5cc53059fb28406c7bc6e7aa/flask_security/utils.py#L276) .

Possible Solution:
----------------------
- Create a view (add via blueprint in the local app?)
```python
@blueprint.route('/cern/logout')
    def logout():
        """Logout view."""
        return redirect("https://login.cern.ch/adfs/ls/partiallogout.aspx", code=302)
```

And set up ``SECURITY_POST_LOGOUT_VIEW`` to this route.

Behaviour:
-------------
- The "Clean Cookie" website is shown. However the cookie is not properly cleant.

PR #182, opened 2018-12-05
oauth: Create CERN OAuth blueprint to allow redirection to login.cern…

ppanero:
Closes https://github.com/inveniosoftware/invenio-oauthclient/issues/180

In order to use it the blueprint should be added:
```python
'invenio_base.blueprints': [
            'cern_oauth = invenio_oauthclient.contrib.cern:cern_logout_blueprint',
        ]
```
And the Flask Security config variable should be set up:
```python
SECURITY_POST_LOGOUT_VIEW="/cern/logout"
```

### DIFF
--- a/invenio_oauthclient/contrib/cern.py
+++ b/invenio_oauthclient/contrib/cern.py
@@ -76,7 +76,7 @@ import copy
 import re
 from datetime import datetime, timedelta
 
-from flask import current_app, g, redirect, session, url_for
+from flask import Blueprint, current_app, g, redirect, session, url_for
 from flask_login import current_user
 from flask_principal import AnonymousIdentity, RoleNeed, UserNeed, \
     identity_changed, identity_loaded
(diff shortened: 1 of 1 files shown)


invenio_oauthclient/contrib/cern.py
```diff
@@ -163,6 +163,15 @@
 REMOTE_APP_RESOURCE_API_URL = 'https://oauthresource.web.cern.ch/api/Me'
 REMOTE_APP_RESOURCE_SCHEMA = 'http://schemas.xmlsoap.org/claims/'
 
+cern_oauth_blueprint = Blueprint('cern_oauth', __name__)
+
+
+@cern_oauth_blueprint.route('/cern/logout')
+def logout():
+    """CERN logout view."""
+    return redirect("https://login.cern.ch/adfs/ls/partiallogout.aspx",
```
ntarocco: I think it would be nice to have this url configurable... WDYT?
You can put it in `REMOTE_APP` config maybe.

kprzerwa: I also think the logout url should be in config variable

PR #186, opened 2019-02-26
utils: fixes missing args from redirect target

Glignos:


invenio_oauthclient/utils.py
```diff
@@ -157,7 +157,11 @@ def get_safe_redirect_target(arg='next'):
             if redirect_uri.host in allowed_hosts:
                 return target
             elif redirect_uri.path:
-                return redirect_uri.path
+                return uricompose(
```
ntarocco: Nice this `uricompose`, I have learnt something! :)
👍

PR #202, opened 2020-04-01
handlers: quick fix for remote api handler when remote is string

pazembrz:
https://its.cern.ch/jira/browse/INSPIR-3317

### DIFF
--- a/invenio_oauthclient/handlers/rest.py
+++ b/invenio_oauthclient/handlers/rest.py
@@ -24,7 +24,8 @@ from ..models import RemoteAccount
 from ..proxies import current_oauthclient
 from ..signals import account_info_received, account_setup_committed, \
     account_setup_received
-from ..utils import oauth_authenticate, oauth_get_user, rest_oauth_register
+from ..utils import oauth_authenticate, oauth_get_user, rest_oauth_register, \
+    obj_or_import_string
 from .utils import get_session_next_url, response_token_setter, token_getter, \
     token_session_key, token_setter
 
@@ -39,6 +40,11 @@ def response_handler_postmessage(remote, url, payload=None):
 
 def default_response_handler(remote, url, payload=None):
     """Default response handler."""
+    default_handler = current_app.config.get(
+        "OAUTHCLIENT_REST_DEFAULT_RESPONSE_HANDLER"
+    )
+    if default_handler:
+        return obj_or_import_string(default_handler)(remote, url, payload)
     scheme, netloc, path, query, fragment = urlsplit(url)
     qs = parse_qs(query)
     if payload:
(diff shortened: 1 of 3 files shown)


invenio_oauthclient/handlers/rest.py
```diff
@@ -50,7 +50,7 @@ def default_response_handler(remote, url, payload=None):
 
 def response_handler(remote, url, payload=None):
     """Handle oauthclient rest response."""
-    if not remote:
+    if not remote or not hasattr(remote, 'name'):
         return default_response_handler(remote, url, payload)
     return current_oauthclient.response_handler[remote.name](
```
drjova, line 55: the issue is also here ``current_oauthclient.response_handler[remote.name]``

invenio_oauthclient/handlers/rest.py
```diff
@@ -39,6 +40,13 @@ def response_handler_postmessage(remote, url, payload=None):
 
 def default_response_handler(remote, url, payload=None):
     """Default response handler."""
+    default_handler = current_app.config.get("OAUTHCLIENT_REST_DEFAULT_RESPONSE_HANDLER")
```
drjova, line 43: we can reuse this https://github.com/inveniosoftware/invenio-oauthclient/blob/aee04536c2c8d1074da227026756074f823c8a81/invenio_oauthclient/utils.py#L168

pazembrz: :+1:

invenio_oauthclient/handlers/rest.py
```diff
@@ -39,6 +39,8 @@ def response_handler_postmessage(remote, url, payload=None):
...
+    return obj_or_import_string(default_handler)(remote, url, payload)
```
drjova, line 43: :/ this return is in the middle of the function

invenio_oauthclient/handlers/rest.py
```diff
@@ -50,7 +52,7 @@ def default_response_handler(remote, url, payload=None):
 
 def response_handler(remote, url, payload=None):
     """Handle oauthclient rest response."""
-    if not remote:
+    if not remote or not hasattr(remote, 'name'):
```
drjova, line 55: we don't need this `` or not hasattr(remote, 'name')``

pazembrz: Then exception will still happen then, as remote is string in our situation and string will be true there

invenio_oauthclient/handlers/rest.py
pazembrz: Argh, I removed one line too much. Sorry I'm tired today.

invenio_oauthclient/handlers/rest.py
drjova: There is no use of ``remote.name``  in ``default_response_handler``, the exception is not happenening anymore becuase the code has been changed since we create the issue. The only part that we have to change is to configure our own ``default_response_handler``.

drjova:
@zzacharo in case of an error currently it's ignoring if you want to use the a different handler than the default one

zzacharo:
@drjova yes in case of error there was no remote at all so we had to make some changes. Personally, I would go probably for a global default error handler so it can be used in the views where no remote is available and thus you cannot use the `response_handler` factory function. Then, I would remove the `if remote...`  check from the response_handler as this was needed only to handle errors in the rest views. WDYT? Also bear in mind that @lnielsen will most probably work on this and refactor a bit the code but certainly include all the features we had agreed during our sprint :)

drjova:
@zzacharo 👍  global default error handler works for us as well. It is actually how @pazembrz has implemented it, we can refactor a bit the PR to reflect this.

drjova:
@zzacharo we've updated the PR

pazembrz:
Ok, I had to add separate error_handler for response_handler as response_handler is called directly from views.
@zzacharo  @drjova

ppanero:
Closing in favour of #205

Issue #203, opened 2020-04-15
Bug on query params in redirect target


carantunes:
Related to: https://github.com/inveniosoftware/invenio-oauthclient/commit/2500dc6935738107617aeade79e050d7608004bb

`uricompose` internally encodes, but accessing query directly with `redirect_uri.query` does not decode, leading to bugs bc of encoding an encoded parameter. 

Ie., I'm getting later on `oauth/errors?error=invalid_request&error_description=Invalid+redirect+URI`

PR #204, opened 2020-04-15
utils: decode redirect target

carantunes:
Fixes: https://github.com/inveniosoftware/invenio-oauthclient/issues/203

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@
 Changes
 =======
 
+Version 1.2.1 (released 2020-04-17)
+
+- Fix args from redirect target' encoding
+
 Version 1.2.0 (released 2020-03-13)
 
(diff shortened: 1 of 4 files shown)


carantunes:
Previously 
`/oauth/authorize?redirect_uri=http%3A%2F%2F127.0.0.1%3A5100%2Fauthorized` 
was being transformed into 
`oauth/authorize?redirect_uri=http%253A%252F%252F127.0.0.1%253A5100%252Fauthorized`

Should I commit the patch version commit into this PR as well?

slint:
Thanks for the example, indeed it's super weird that it got serialized like this...

> Should I commit the patch version commit into this PR as well?

Yes, but put it in a separate commit with a `release: v1.2.1` message (like the one I linked for `v1.2.0`).

carantunes:
Done @slint

PR #210, opened 2020-06-03
handlers: export rest handlers

ppanero:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/188

- Ideal would be to rename ui exports (import as ....) but that would break compatibility with 1.3.0 release
- How could a test for this be made? Sadly it fails only at run time but without a IdP this cannot be checked...

Import works:
Screenshot 2020-06-03 at 13 33 31

**RDM needs this released**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@
 Changes
 =======
 
+Version 1.3.1 (released 2020-06-03)
+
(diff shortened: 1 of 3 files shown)


ppanero:
@slint done, check if the changelog seems correct, after is ready for release :)

PR #221, opened 2020-09-17
global: clear roles and cern_resource on logout

### LINKED ISSUE #220: CERN contribs do not clear `provides` on logout
Package version (if known): any

## Describe the bug
When logging in, the `extend_identity` method is adding some `roles` to the identity. When logging out, such `roles` are not cleaned.
Probably this cleaning should be done [here](https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/contrib/cern.py#L507), before `return`.

Logged in user:

## Current behaviour

Never logged in user:

```
<AnonymousIdentity id="None" auth_type="None" provides={Need(method='system_role', value='any_user')}>
```

Authenticated user:

```
<Identity id="7" auth_type="None" provides={Need(method='id', value=7), Need(method='role', value='cern-user'), Need(method='system_role', value='authenticated_user'), Need(method='id', value='<email>@cern.ch'), Need(method='system_role', value='any_user'), Need(method='role', value='admin')}>
```

Anonymous user after logout:

```
<AnonymousIdentity id="None" auth_type="None" provides={Need(method='id', value='<email>@cern.ch'), Need(method='system_role', value='any_user'), Need(method='role', value='admin'), Need(method='role', value='cern-user')}>
```

## Expected behaviour

Anonymous user after logout:

```
<AnonymousIdentity id="None" auth_type="None" provides={Need(method='system_role', value='any_user')}>
```

topless:
* removes `cern_resource` from session when user logs out
* removes `roles` from session when user logs out
* closes #220


zzacharo (approved):

invenio_oauthclient/contrib/cern.py
```diff
@@ -504,6 +505,7 @@ def on_identity_changed(sender, identity):
     :param identity: The user identity where information are stored.
     """
     if isinstance(identity, AnonymousIdentity):
+        disconnect_identity(identity)
```
kprzerwa, line 508: IMO this is "too late" to clean the global identity - it is already the AnonymousIdentity so it should have no leftovers at this stage. Can't we connect to logout signal somehow and clean it there?

topless: I think this is good thinking here.

These signals are from flask principal, while flask-login provides the explicit [signals](https://flask-login.readthedocs.io/en/latest/#signals) login, logout etc.. you refer to.

`invenio-accounts` comes with a [listener](https://github.com/inveniosoftware/invenio-accounts/blob/b156d3bcabba2b2c4008b4c023a36d44f87888e7/invenio_accounts/sessions.py#L93) that is cleaning some stuff in the session. I believe that those who declare some props in the session should also be responsible for cleaning them up, i.e. `cern_resource` value. The `identity.provides` are part of the principal module, thus their `reset` in that signal.

I do not have a strong opinion on it, maybe @zzacharo can advice on this.

ntarocco: I don't think this is too late. This signal is fired when the identify changes, from Authenticated -> Anonymous, so it should be correct.
Let's discuss IRL if needed.

invenio_oauthclient/contrib/cern_openid.py
```diff
@@ -210,6 +210,7 @@ def extend_identity(identity, roles):
 
 def disconnect_identity(identity):
     """Disconnect identity from CERN groups."""
+    session.pop("cern_resource", None)
```
ntarocco, line 213: I find the name `disconnect_identity` quite misleading... but probably for another time

ntarocco (approved):

invenio_oauthclient/contrib/cern_openid.py
topless: `disconnect_identity` the disconnect term is used in a more wide spectrum in the repo, not just in this case, thus I let it be. I agree that it might not be the most self explained.

zzacharo: I guess it was used because until now was part of the "disconnect" workflow :)

tests/test_contrib_cern.py
```diff
@@ -194,3 +205,15 @@ def test_account_info_not_allowed_account(app, example_cern):
     assert resp.status_code == 302
     assert session['_flashes'][0][0] == 'danger'
     assert session['_flashes'][0][1] == 'CERN account not allowed.'
+
+
+def test_empty_disconnect_identity(app):
```
ntarocco, line 210: @topless I am not sure I understand this test, what is its purpose?

tests/test_contrib_cern.py
```diff
@@ -138,6 +139,16 @@ def test_account_setup(app, example_cern, models_fixture):
 
         login_user(user)
         assert len(g.identity.provides) == 7
+
```
ntarocco, line 142: Can you please add something like (pseudo-code):
```
login()
assert type(g.identity) == Identity
assert g.identity.provides == Need(method='role', value='cern-user'), Need(method='system_role', value='authenticated_user'), Need(method='id', value='<email>@cern.ch'), Need(method='system_role', value='any_user'), Need(method='role', value='admin')

logout()
assert type(g.identity) == AnonymousIdentity
assert g.identity.provides == Need(method='system_role', value='any_user')

login()
...test again...
```
as in the GitHub issue.
Same for the other tests.

tests/test_contrib_cern.py
topless: The initial problem was that when there was no value in AnonymousIdentity.provides, we were returning a dict instead of a set. Tests didn't catch it because there was always an entry in provides. The purpose of this test was to validate, that we pop the correct type from the session when there was nothing there. IMO is redundant especially being replicated every time for every provider and for both ui and rest.

tests/test_contrib_cern.py
topless: To follow our morning catch up, when we log out the `AnonymousIdentity` provides are `UserNeed({userId})`, instead of the generic `RoleNeed('any_user')`. I have left a [note](https://github.com/topless/invenio-oauthclient/blob/05cd1be4f99b499241b3a4bd3191b5cf22763237/tests/test_contrib_cern_rest.py#L145) in one of the tests.

pytest.ini
```diff
@@ -7,7 +7,6 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 [pytest]
-pep8ignore = docs/conf.py ALL
-addopts = --pep8 --doctest-glob="*.rst" --doctest-modules --cov=invenio_oauthclient --cov-report=term-missing
+addopts = --isort --pydocstyle --pycodestyle --doctest-glob="*.rst" --doctest-modules --cov=invenio_oauthclient --cov-report=term-missing
 testpaths = docs tests invenio_oauthclient
-filterwarnings = ignore::pytest.PytestDeprecationWarning
\ No newline at end of file
+filterwarnings = ignore::pytest.PytestDeprecationWarning
```
ntarocco, line 12: this was removed I believe

ntarocco (approved):

pytest.ini
ntarocco: You might also have to change `run-tests.sh`?

topless: It was suggested by @lnielsen in https://github.com/inveniosoftware/cookiecutter-invenio-module/pull/135/files

PR #224, opened 2020-10-02
global: clear roles and cern_resource on logout

ntarocco:


ntarocco:
We need to add fixes for cern contribs in both master and `maint-v1.3` branches

PR #225, opened 2020-10-13
contrib: handle login via token

kprzerwa:


invenio_oauthclient/contrib/cern_openid.py
```diff
@@ -370,26 +372,34 @@ def on_identity_changed(sender, identity):
...
+    if remote_account and not logged_in_via_token:
         refresh = current_app.config.get(
             "OAUTHCLIENT_CERN_OPENID_REFRESH_TIMEDELTA",
             OAUTHCLIENT_CERN_OPENID_REFRESH_TIMEDELTA,
         )
-
-        roles.extend(
-            account_roles_and_extra_data(
-                account, resource, refresh_timedelta=refresh
+        if refresh:
```
ntarocco, line 391: Do you need this `if refresh`? I think it will be always set?

invenio_oauthclient/contrib/cern_openid.py
```diff
@@ -370,26 +372,34 @@ def on_identity_changed(sender, identity):
...
+            remote = find_remote_by_client_id(client_id)
+            resource = get_resource(remote)
+            roles.extend(
+                account_roles_and_extra_data(
+                    remote_account, resource, refresh_timedelta=refresh
+                )
             )
-        )
+        else:
+            roles.extend(remote_account.extra_data["roles"])
```
ntarocco, line 400: it might be not set yet?

ntarocco (commented):
Can you add the same for `cern` contrib?

invenio_oauthclient/contrib/cern_openid.py
kprzerwa: No, If you set a False value on the config the refresh will be turned off, I did it to leave the functionality of refresh, I thought it might be handy for some use cases

kprzerwa:
> Can you add the same for `cern` contrib?

@ntarocco shouldn't I add the deprecation warning instead?

tests/test_contrib_cern_openid_rest.py
```diff
@@ -196,3 +199,85 @@ def test_account_info_not_allowed_account(app_rest, example_cern_openid_rest):
         "code": 400,
     }
     check_response_redirect_url_args(resp, expected_url_args)
+
+
+def test_identity_changed(app_rest, example_cern_openid_rest, models_fixture):
```
ntarocco, line 204: Wow... not easy to understand this test... a few tips:
* could you maybe wrap in a function the init/preparation of the test, e.g.
```
def test():
    def _init(): 
         ....
    init()
    ...real test code...
```
* I think that the main thing would be to test if the extra metadata has been fetched or not. I think that I would expect to see:

```
def test():
    ...
    def _test_login_token_no_remote_account():
        assert login succeed
        assert no extra metadata

     def _test_login_token_with_remote_account():
        assert login succeed
        assert extra metadata
```

kprzerwa: fixed

ntarocco (approved):

ntarocco:
@kprzerwa go ahead and release if you want. Do not forget to cherry-pick this commit in `master` branch!

PR #226, opened 2020-10-15
contrib: logout redirection to CERN auth

kprzerwa:


invenio_oauthclient/contrib/cern_openid.py
```diff
@@ -402,3 +409,21 @@ def on_identity_loaded(sender, identity):
...
+
+    def get(self):
+        """Logout a user."""
+        return self.logout_user()
+
+
+@cern_oauth_blueprint.route('/cern_openid/logout/')
+def logout_redirect():
+    """Redirect to the CERN OpenID logout."""
+    return redirect(current_app.config['OAUTH_REMOTE_REST_APP']['logout_url'])
```
topless, line 429: What is the code that will return?
```suggestion
    return redirect(current_app.config['OAUTH_REMOTE_REST_APP']['logout_url'], code=302)
```

invenio_oauthclient/contrib/cern_openid.py
```diff
@@ -90,20 +94,24 @@
 OAUTHCLIENT_CERN_OPENID_ALLOWED_ROLES = ["cern_user"]
 """CERN OAuth application role values that are allowed to be used."""
 
+_ACCOUNTS_REST_AUTH_VIEWS.update(
+    logout="invenio_oauthclient.contrib.cern_openid:CERNOpenIDLogoutView")
+ACCOUNTS_REST_AUTH_VIEWS = _ACCOUNTS_REST_AUTH_VIEWS
```
topless, line 99: Can we just skip the whole `_ACCOUNTS_REST_AUTH_VIEWS`? I think it is just an alias, so we could just 

```
from invenio_accounts.config import ACCOUNTS_REST_AUTH_VIEWS
    
ACCOUNTS_REST_AUTH_VIEWS.update(
    logout="invenio_oauthclient.contrib.cern_openid:CERNOpenIDLogoutView")
```

topless (approved):

invenio_oauthclient/contrib/cern_openid.py
kprzerwa: I had problems with updating the config variables in this way in the past, they were not overwritten correctly

ntarocco: There might be a problem here: it looks like that `ACCOUNTS_REST_AUTH_VIEWS` is global. So if you have configured multiple login method, e.g. local accounts, GitHub and CERN, they will all use this class.
@zzacharo can you maybe give your advice?
No idea how to solve it at the moment...

zzacharo: Why don't you decouple this behaviour in the UI. For example:

- SPA logout from accounts
- SPA logout from OAuth

invenio_oauthclient/contrib/cern_openid.py
zzacharo: I think is the case already when you return the `redirect` :)

zzacharo, line 429: @kprzerwa So, after the redirect to the logout page of OAuth, are you staying on that page, i.e not returning back to the SPA?

zzacharo: @topless Actually, you might be right 😅

ntarocco: It is `302` by default... but yeah better safe than sorry :)

PR #231, opened 2020-12-01
ci: migrate to GA

### LINKED ISSUE #230: global: migration to github-actions from travis

mvidalgarcia:
closes #230


run-tests.sh
```diff
@@ -7,7 +7,19 @@
...
-sphinx-build -qnNW docs docs/_build/html && \
+# Quit on errors
+set -o errexit
+
+# Quit on unbound symbols
+set -o nounset
+
+python -m check_manifest --ignore ".*-requirements.txt"
+python -m sphinx.cmd.build -qnNW docs docs/_build/html
+docker-services-cli up ${DB}
```
diegodelemos, line 21: Minor: FYI @lnielsen, [according to the old `.travis.yml`](https://github.com/inveniosoftware/invenio-oauthclient/blob/master/.travis.yml#L18) we needed Redis so this should be `docker-services-cli up redis ${DB}`, but actually all the whole build passes without it...

mvidalgarcia: Wow, I hadn't realized at all. Good question.

diegodelemos (approved):

run-tests.sh
lnielsen: This is probably a mistake in the old Travis file......I'm trying to remember why oauthclient would need it.

PR #242, opened 2021-02-05
contrib: fix retrieve user info in CERN OpenID

ntarocco:
* makes /userinfo endpoint configurable
* makes params passed to the JWT decode function configurable


ntarocco:
We will add this commit only on `master`

PR #246, opened 2021-05-04
contrib: refactor OAuth to add settings helper

ntarocco:


invenio_oauthclient/contrib/keycloak/helpers.py
```diff
@@ -46,22 +45,25 @@ def get_user_info(remote, resp_token,
         endpoint mechanism when verifying the 'id_token' fails
     :param options: A dictionary with additional options for `jwt.decode`
     """
+    app_name = remote.name
+    config_prefix = f"OAUTHCLIENT_{app_name.upper()}"
```
max-moser, line 49: should we maybe add some sanitation to `app_name`?
i'm thinking because it could be harder to configure via `invenio.cfg`, if the name contains characters that are invalid for python variable names, e.g. whitespaces or dashes

i think it could still be configured via `app.config["OAUTHCLIENT_MY-KEYCLOAK_AUD"]`, etc., but it would be a bit harder to do
the trade-off would likely be that we have to precisely document this behaviour (something like https://click.palletsprojects.com/en/7.x/options/#name-your-options)

max-moser (approved):
that looks great!
thanks, this should make the configuration quite a bit more intuitive :)

invenio_oauthclient/contrib/globus.py
```diff
@@ -43,72 +43,111 @@
...
+            authorized_redirect_url='/',
+            disconnect_redirect_url='/',
+            signup_redirect_url='/',
+            error_redirect_url='/'
+        )
 
+    @property
+    def user_info_url(self):
+        """Return the URL to fetch user info."""
+        return f"{self.base_url}oauth2/userinfo"
```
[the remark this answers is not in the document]
ntarocco: Yes, you are right, forgot to check that. I am removing Python 2 and going for Python 3.6 min.

invenio_oauthclient/contrib/keycloak/helpers.py
ntarocco: That's why I don't like these hidden conventions... I think you are right, I will add some checks

ntarocco:
@max-moser I will merge it, please ping me if you want to change anything, I can create another PR ;)

PR #247, opened 2021-05-05
login: add possibility to auto-redirect to external auth services

max-moser:
partially closes https://github.com/inveniosoftware/docs-invenio-rdm/issues/153
requires https://github.com/inveniosoftware/invenio-accounts/pull/368

TODO:
the provided config for `ACCOUNTS_LOGIN_VIEW_FUNCTION` should override the value from `invenio_accounts.config`, but doesn't

### DIFF
--- a/invenio_oauthclient/views/client.py
+++ b/invenio_oauthclient/views/client.py
@@ -8,8 +8,9 @@
 
 """Client blueprint used to handle OAuth callbacks."""
 
-from flask import Blueprint, abort, current_app, request, url_for
+from flask import Blueprint, abort, current_app, redirect, request, url_for
 from flask_oauthlib.client import OAuthException
+from invenio_accounts.views import login as base_login
 from invenio_db import db
 from itsdangerous import BadData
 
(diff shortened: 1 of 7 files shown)


invenio_oauthclient/views/client.py
```diff
@@ -64,6 +65,28 @@ def post_ext_init(state):
...
+    remote_apps = list(current_oauthclient.oauth.remote_apps)
+
+    if would_redirect and len(remote_apps) == 1:
+        # if local login is disabled and we only have one OAuth2 remote app
+        # configured, we forward directly to that
+        url = url_for("invenio_oauthclient.login", remote_app=remote_apps[0])
+        return redirect(url)
+
+    else:
+        return base_login(*args, **kwargs)
```
ntarocco, line 87: There might be an issue. Imagine the scenario:
- I have configured for `invenio-accounts` `ACCOUNTS_LOGIN_VIEW_FUNCTION = myloginfunc`
- I install `invenio-oauthclient`, and it overrides it to `ACCOUNTS_LOGIN_VIEW_FUNCTION = auto_redirect_login`
- if I have `ACCOUNTS_LOCAL_LOGIN_ENABLED = True`, it will return `base_login` and not `myloginfunc`

or maybe this is not an issue because in this case I will have to redefine `auto_redirect_login` to return`myloginfunc` ...

max-moser: hmm... wouldn't the configured value (via `invenio.cfg`, i suppose?) then still take precedence over the default values provided by any of the modules?

ntarocco (approved):

setup.py
```diff
@@ -61,13 +62,14 @@
 ]
 
 install_requires = [
+    'invenio-base>=1.2.4',
+    'Flask>=1.1.4,<2.0.0',
     'Flask-Breadcrumbs>=0.5.0',
     'requests-oauthlib>=0.6.2,<1.2.0',
     'oauthlib>=1.1.2,<3.0.0',
-    'Flask-OAuthlib>=0.9.5',
+    'Flask-OAuthlib>=0.9.6',
```
max-moser, line 70: this particular bump was required to not have the autodoc fail

ntarocco (approved):

PR #252, opened 2021-05-11
settings: disable unlinking of last external account

### LINKED ISSUE #251: Allow disabling of "disconnect" button for linked accounts
## Is your feature request related to a problem? Please describe.
The PR https://github.com/inveniosoftware/invenio-accounts/pull/368 allows disabling of logging in using "local" accounts, leaving only external authentication services as means for authentication.
If every linked account would be disconnected, it *might* be possible to create a situation where there's no easy way of logging into that account anymore (would probably involve changing the registered e-mail address of the user and disconnecting *all* linked accounts).

## Describe the solution you'd like
There should be some way to prevent the users from accidentally locking themselves out of the account that easily.
Probably this would include disabling the "disconnect" button for the last linked external account in case that local logins are disabled, although the precise conditions still have to be discussed.

max-moser:
* this only applies if the user can't login to their local account
  (either ACCOUNTS_LOCAL_LOGIN_ENABLED=False, or they don't have a
  password set)

closes #251

### DIFF
--- a/invenio_oauthclient/handlers/utils.py
+++ b/invenio_oauthclient/handlers/utils.py
@@ -11,14 +11,14 @@
 from functools import partial, wraps
 
 import six
-from flask import current_app, session
+from flask import abort, current_app, session
 from flask_login import current_user
 from invenio_db import db
 from werkzeug.utils import import_string
 
 from ..errors import OAuthClientError, OAuthRejectedRequestError, \
     OAuthResponseError
-from ..models import RemoteToken
+from ..models import RemoteAccount, RemoteToken
 from ..proxies import current_oauthclient
 
 
@@ -245,4 +245,47 @@ def authorized_handler(f, authorized_response):
     def decorated(*args, **kwargs):
         data = authorized_response()
         return f(*((data,) + args), **kwargs)
+    return decorated
+
+
+def require_more_than_one_external_account(f):
+    """Require that the user has more than one external account for login.
+
+    If the user only has one linked external account and no means for logging
+    in via local credentials, the decorated function won't be executed.
+    This decorator is useful for disconnect handlers, to prevent users from
+    disconnecting their last potential means of authentication.
+    """
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        if current_user.is_anonymous:
+            return f(*args, **kwargs)
+
+        local_login_enabled = current_app.config.get(
+            "ACCOUNTS_LOCAL_LOGIN_ENABLED", True
+        )
+        password_set = current_user.password is not None
+        local_login_possible = local_login_enabled and password_set
+
+        remote_apps = current_app.config["OAUTHCLIENT_REMOTE_APPS"]
+        accounts = RemoteAccount.query.filter_by(
+            user_id=current_user.get_id()
+        ).all()
+
+        # find out all of the linked external accounts for the user
+        # that are currently configured and not hidden
+        linked_accounts = [
+            acc for acc in accounts
+            if acc.client_id in remote_apps and
+            not remote_apps[acc.client_id].get("hide", False)
+        ]
+
+        # execute the function only if local login is possible, or
+        # there's more than one linked external account
+        if local_login_possible or len(linked_accounts) > 1:
+            return f(*args, **kwargs)
+
+        else:
+            abort(400)
+
     return decorated
--- a/invenio_oauthclient/templates/invenio_oauthclient/settings/index.html
+++ b/invenio_oauthclient/templates/invenio_oauthclient/settings/index.html
@@ -14,6 +14,7 @@
 {% set panel_title = _("Linked accounts") %}
 {% set panel_icon = "fa fa-link" %}
 {% set num_linked_services = services|selectattr("account")|list|length %}
+{% set can_disconnect = (num_linked_services > 1 or not only_external_login) %}
 
 {% block settings_body %}
 <div class="panel-body">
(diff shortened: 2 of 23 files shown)


invenio_oauthclient/views/settings.py
```diff
@@ -79,7 +79,15 @@ def index():
...
+        "ACCOUNTS_LOCAL_LOGIN_ENABLED", True
+    )
+    password_set = current_user.password is not None
+    local_login_possible = local_login_enabled and password_set
+
     return render_template(
         'invenio_oauthclient/settings/index.html',
-        services=services
+        services=services,
+        local_login_impossible=not local_login_possible,
```
max-moser, line 92: i've added this kind of inverted logic here because a falsy value for this parameter (e.g. by leaving it out) would produce a result that's most in line with how everything worked before (no restriction of rendering on any "disconnect" buttons)

ntarocco: Agree, I would maybe have changed the name of the var to something like `only_external_login`.

invenio_oauthclient/templates/invenio_oauthclient/settings/index.html
```diff
@@ -13,6 +13,7 @@
 
 {% set panel_title = _("Linked accounts") %}
 {% set panel_icon = "fa fa-link" %}
+{% set num_linked_services = services|selectattr("account")|list|length %}
```
ntarocco, line 16: I am not sure we need the check for local login. Isn't it enough to check only if there is one or multiple external accounts?
Minor suggestion:
```suggestion
{% set num_linked_services = services|selectattr("account")|list|length %}
{% set can_disconnect = num_linked_services > 1 or not local_login_impossible %}

....

{%- if s.account and can_disconnect -%}
```

ntarocco (commented):
Looks good, thank you!

Does it make sense maybe to also block the `disconnect` handler, even if it might be overridden by a contrib?

```
def _disconnect(remote, *args, **kwargs):
    """Common logic for handling disconnection of remote accounts."""
    if not current_user.is_authenticated:
        return current_app.login_manager.unauthorized()
    
    if only_one_account:
        abort(...)
```

invenio_oauthclient/templates/invenio_oauthclient/settings/index.html
max-moser: do you mean we should always disable the disconnect of the last external auth, regardless of whether or not a local login is possible?

ntarocco: If there is only 1 external auth, then maybe the button should be disabled independently of the local login. Or am I missing something?

max-moser: sure, we can ignore the possibility for local login and just prohibit disconnecting the last remaining external auth anyway
that would probably render some of the changes in this PR irrelevant, and the code could be slightly simplified again
i'll ask again in the auth channel in discord, to make sure we're all on the same page

setup.py
```diff
@@ -61,13 +61,14 @@
 ]
 
 install_requires = [
+    'invenio-base>=1.2.4',
+    'Flask>=1.1.0,<2.0.0',
```
max-moser, line 65: this is "required" because the pip resolver takes ages to resolve the correct set of versions of all packages otherwise

max-moser:
of course, we could still make this behaviour optional by adding yet another configuration item, and checking for it in the relevant places

invenio_oauthclient/handlers/utils.py
```diff
@@ -246,3 +246,43 @@ def decorated(*args, **kwargs):
...
+        accounts = RemoteAccount.query.filter_by(
+            user_id=current_user.get_id()
+        ).all()
+
+        # find out all of the linked external accounts for the user
+        # that are currently configured and not hidden
+        linked_accounts = [
+            acc for acc in accounts
+            if acc.client_id in remote_apps and
+            not remote_apps[acc.client_id].get("hide", False)
```
ntarocco, line 280: I was not aware of the `hide` feature, learned something today! :)

ntarocco, line 280: Question: is it possible that `emote_apps[acc.client_id]` fails?
I am thinking to this very very very uncommon scenario:
1. sysadmin configure remote app `A`
2. users login with `A`
3. after something, institutions change to remote app `B`, and remove `A` from configured remote apps
4. users login with `B` and go to the user profile, click on `Disconnect` on `A`

It might fail? But also the disconnection will fail... might be a very weird use case...

invenio_oauthclient/templates/invenio_oauthclient/settings/index.html
```diff
@@ -13,6 +13,8 @@
...
+{% set can_disconnect = (num_linked_services > 1 or not local_login_impossible) %}
```
ntarocco, line 17: The name of the var has changed to `only_external_login` if I am not mistaken:
```suggestion
{% set can_disconnect = (only_external_login and num_linked_services > 1) %}
```

The same below.

invenio_oauthclient/handlers/utils.py
max-moser: me neither, i just stumbled upon it in the code: https://github.com/inveniosoftware/invenio-oauthclient/blob/f359dbddf7a35eee22af6510eb29241938402c5b/invenio_oauthclient/views/settings.py#L59

invenio_oauthclient/templates/invenio_oauthclient/settings/index.html
max-moser: indeed, nice catch!

invenio_oauthclient/handlers/utils.py
max-moser: that shouldn't be the case here, due to the filtering `if acc.client_id in remote_apps`, but i also have the feeling that this case could potentially lead to some unexpected situations generally

ntarocco (approved):

ntarocco (approved):

PR #254, opened 2021-05-25
contrib: fix CERN contrib signal

### LINKED ISSUE #228: Login/logout should know the method used
Currently, CERN contribs (and probably others) relies on signals to perform login/logout operations.
Unfortunately, all the contribs are "registered" on app init and they will all receive signals. 

### Problem

On login/logout, each contribs receive the signal and perform operations. The correct behaviour would be that only the contrib used to login should receive such signals.
This is also creating an issue on logout: given that `invenio-oauthclient` is not aware of which login method was used, the logout signal is emitted to all contribs and they all perform logout operations, causing issues in case of redirections.

### Possible solution

The solution could be to use DB information or add any extra information to store which was the authentication method used by the user. This could allow to stop firing signals to all contribs and emit only to the contrib used on login.

ntarocco:
* skip CERN contrib signals when login happened via another login method
* closes #228

### DIFF
--- a/invenio_oauthclient/contrib/cern.py
+++ b/invenio_oauthclient/contrib/cern.py
@@ -391,6 +391,7 @@ def get_resource(remote):
 
 def _account_info(remote, resp):
     """Retrieve remote account information used to find local user."""
+    g.oauth_logged_in_with_remote = remote
     resource = get_resource(remote)
 
     valid_identities = current_app.config.get(
@@ -511,6 +512,11 @@ def on_identity_changed(sender, identity):
         disconnect_identity(identity)
         return
 
+    remote = g.get("oauth_logged_in_with_remote", None)
+    if not remote or remote.name != "cern":
+        # signal coming from another remote app
+        return
+
     client_id = current_app.config['CERN_APP_CREDENTIALS']['consumer_key']
     account = RemoteAccount.get(
         user_id=current_user.get_id(),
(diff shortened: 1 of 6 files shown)


ntarocco:
@GraemeWatt @alisonrclarke are you able to test this PR? From my tests it fixes the issue.

GraemeWatt:
I've tested [your new branch](https://github.com/ntarocco/invenio-oauthclient/tree/fix-login-signal-cern-contribs) with the [HEPData web app code](https://github.com/HEPData/hepdata) and I confirm that it fixes issue #228.  I no longer get the `KeyError: 'cern_roles'` exception from local login.  Many thanks for providing a solution so quickly.  Will you make a new release of the `invenio-oauthclient` package after merging this PR?

kprzerwa (approved):

ntarocco:
> I've tested [your new branch](https://github.com/ntarocco/invenio-oauthclient/tree/fix-login-signal-cern-contribs) with the [HEPData web app code](https://github.com/HEPData/hepdata) and I confirm that it fixes issue #228. I no longer get the `KeyError: 'cern_roles'` exception from local login. Many thanks for providing a solution so quickly. Will you make a new release of the `invenio-oauthclient` package after merging this PR?

Yes indeed, I am just doing the last tests on the the module, I should be able to release it today.

Issue #267, opened 2021-12-15
Faulty logic in disconnect external account check


slint:
Package version (if known): v1.5.3

## Describe the bug

Even though someone might have more than one external accounts connected, the check performed in [`require_more_than_one_external_account`](https://github.com/inveniosoftware/invenio-oauthclient/blob/9b8bd7bc8bcbbe178aad3f0f8a2e620749c9980b/invenio_oauthclient/handlers/utils.py#L279), is faulty in terms of how client configuration is usually advertised in the documentation. 

Usually, configurations are like:

```python
OAUTHCLIENT_REMOTE_APPS = dict(
    github=github.REMOTE_APP,
    orcid=orcid.REMOTE_APP,
)
GITHUB_APP_CREDENTIALS = dict(
    consumer_key="abc123",
    consumer_secret="...",
)
ORCID_APP_CREDENTIALS = dict(
    consumer_key="def456",
    consumer_secret="...",
)
```

[In the code](https://github.com/inveniosoftware/invenio-oauthclient/blob/9b8bd7bc8bcbbe178aad3f0f8a2e620749c9980b/invenio_oauthclient/handlers/utils.py#L279), the `acc.client_id` is actually the application's Client ID/Key, e.g. in this example  `abc123` (for GitHub) or `def456` (for ORCiD), and the keys in `remote_app` (which is `OAUTHCLIENT_REMOTE_APPS` config) are `github` and `orcid`.

This means that disconnecting a linked account will always fail since `len(linked_accounts)` will always be 0, even if the user has actually connected more than one external account. 

## Steps to Reproduce

1. Setup two OAuth authentication apps (using the usually proposed config described above).
2. Sign-up with one of the above external methods and connect other
2. (Don't set a password)
3. Go to the linked accounts settings page (e.g. https://localhost:5000/account/settings/linkedaccounts/)
4. Click the "Disconnect" button for any of the two linked accounts
5. Get a "400 Bad Request" response page

## Additional context

For backward compatibility (because for users that want this to work with multiple external accounts, they might have used the actual client IDs as keys in the `OAUTHCLIENT_REMOTE_APPS` config), we might have to check both client ID and remote app config key?

[...]

PR #268, opened 2022-01-06
handlers: fix linked account disconnection

ppanero:
- closes https://github.com/inveniosoftware/invenio-oauthclient/issues/267

@slint are tests needed? I can see for example in the ORCiD tests (`test_authorized_signup`), there is a test for the 400 error done when disconnecting (and then it works). I could test the decorator but not sure we can (because of mocking) test something in a meaningful way for the views.

https://user-images.githubusercontent.com/6756943/148374499-ffb3de72-d8e8-402b-9c1e-2854f75a8ccc.mov


max-moser (approved):
LGTM

ppanero:
@slint shall we merge?

Issue #277, opened 2022-05-18
contrib login: silent failure when username exists 


ppanero:
Package version (if known): v2.0.0.dev1

## Describe the bug

When login with a contrib (e.g. CERN) if the username already exists it asks you to "complete the registration" but without any indication of _username already exists in the system_. Therefore, as a user I would just confirm. Then returns to the homepage without any error message and not login/registering the user.

## Steps to Reproduce
1. register a local user (e.g. your cern or orcid username)
2. try to login with a contrib with the same username (e.g. cern)
3. confirm registration

## Expected behavior

1. The form should display the username field in red2. 
2. The form should not allow confirmation, or fail to confirm if the username is unchanged.3. 

## Screenshots (if applicable)

https://user-images.githubusercontent.com/6756943/169042747-e1a340e4-ecb1-4abb-942a-35ce61d214a9.mov

## Additional context

This problem is being caused by two issues:

When logging in with a contrib, when the user name is already taken the user is shown the [confirm registration/signup form](https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/templates/semantic-ui/invenio_oauthclient/signup.html).

The form will be submitted, without reporting errors (nor returning a `next_url` to redirect). Leaving the `handelrs/ui.py:signup_hanlder` to redirect to the homepage (without displaying any error).

```python
        if form.is_submitted():
            if next_url:
                return redirect(next_url)
            else:
                return redirect('/')
```

On the other hand, at this point (before the if) the form does contain the errors.

```shell
(Pdb) form.errors
{'profile': {'username': [l'Username is not available.']}}
```

The local user registration does not have this bug because the views works in a slightly different way. See [here](https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/views/rest.py#L283)

ppanero:
may be related to https://github.com/inveniosoftware/invenio-oauthclient/issues/259

PR #278, opened 2022-05-20
handlers: report errors on signup

ppanero:
- closes https://github.com/inveniosoftware/invenio-oauthclient/issues/277

The final result with all possible errors triggered:

Screenshot 2022-05-20 at 12 02 03

An ideal and more consistent solution would be similar to the following. However, the problem is that `profile` is a single field (with three inner ones: username, full_name and affiliations). No way to forward the errors from `render_field` to the actual field rendering (L46) was found. The local form (coming form invenio-userprofiles) can do that becuase it renders fields explicitly (i.e. not in a for loop) and the are no field with inner fields in the form.

Screenshot 2022-05-20 at 10 58 22


invenio_oauthclient/templates/invenio_oauthclient/_macros.html
```diff
@@ -18,22 +18,6 @@
 {% endmacro %}
 
 
-{% macro form_errors(form) %}
```
ppanero, line 21: This macro was not used anywhere else except invenio-accounts and invenio-oauthclient (see [organization search](https://github.com/search?p=3&q=org%3Ainveniosoftware+form_errors&type=Code). However, both modules also include the code for it in them, so there is no need to have it here. The same applies to the semanti-ui macro.

ppanero:
# Status

This PR is blocked because the CSRF token is not getting into the form (and therefore it does not validate, and cant login with contribs). Things tested:
- `WTF_CSRF_ENABLED` coming form Falsk-WTF
- Passing the `meta={'csrf: True'}` to the form construction as happens in some examples of Flask-Security
- ...

slint:
In the end, there was probably some copy-paste issue with the Jinja macro, since the line that renders hidden form fields was removed (leaving an "empty" `if` statement clause in Jinja). The CSRF field is now properly rendered and validates when the form is submitted.

PR #288, opened 2022-12-01
global: allow extra settings per app

ntarocco:
* separates the fetching of the user info and the serializaton of its response, adding a hook to provide a custom serializer. Useful when we want to only provide a different serializer.
* add extra signup params to allow to auto-confirm user and optionally send the registration e-mail.
* changes the OAuth apps to automatically auto-confirm users, except for the ORCID app that requires the user to provide a valid e-mail address to be then confirmed.
* deprecates the old CERN OAuth contrib.

Requires https://github.com/inveniosoftware/invenio-accounts/pull/429
Tests should pass when `invenio-accounts` is released.


invenio_oauthclient/contrib/cern_openid.py
```diff
@@ -104,6 +104,10 @@
     icon="",
     logout_url="https://auth.cern.ch/auth/realms/cern/protocol/"
     "openid-connect/logout",
+    signup_options={
```
ntarocco, line 107: Default for `orcid` app

invenio_oauthclient/contrib/settings.py
```diff
@@ -46,11 +47,24 @@ def __init__(
             },
         }
 
+        signup_options = signup_options or {}
+        signup_options.setdefault("auto_confirm", True)
```
ntarocco, line 51: defaults for all OAuth/Keyclack apps

invenio_oauthclient/contrib/settings.py
```diff
@@ -46,11 +47,24 @@ def __init__(
...
+        signup_options.setdefault("send_register_msg", False)
+
+        if (
+            not signup_options["auto_confirm"]
+            and not signup_options["send_register_msg"]
+        ):
+            raise ValueError(
```
ntarocco, line 58: I am not 100% about this check... is it something that we should enforce? Without, it can create issues (user cannot confirm).

tests/test_contrib_cern_openid.py
```diff
@@ -127,7 +128,7 @@ def test_account_setup(app, example_cern_openid, models_fixture):
         g.oauth_logged_in_with_remote = ioc.remote_apps["cern_openid"]
 
         login_user(user)
-        assert len(g.identity.provides) == 3
+        assert len(g.identity.provides) == 2
```
ntarocco, line 131: @zzacharo any idea what change could have caused this? It does not seem related to the changes in this PR.

invenio_oauthclient/contrib/cern_openid.py
alejandromumo: minor: isn't this the default for remote apps set below?

invenio_oauthclient/contrib/settings.py
alejandromumo: In my opinion, if we don't have a way to resend the confirmation e-mail then it's better to have the check than not .
However it's possibly an underlying issue that an user can end up in that situation in the first place. It might even happen that, by some unexpected issue, the e-mail is not sent / received and the user is left in the "void".

alejandromumo (commented):
LGTM: added just minor comments / discussion

invenio_oauthclient/contrib/cern_openid.py
ntarocco: This CERN OpenID is a bit "special", fully custom :( It does not inherit from the OAuth helper as all the others.

max-moser (approved):

invenio_oauthclient/utils.py
```diff
@@ -186,14 +188,21 @@ def oauth_register(form, user_info=None, precedence_mask=None):
...
         # see https://github.com/inveniosoftware/invenio-oauthclient/issues/275
         remove_none_values(data)
         # remove the CSRF tokens to avoid unexpected keyword arguments
         remove_csrf_tokens(data)
-        user = register_user(**data)
+
+        auto_confirm = signup_options.get("auto_confirm", False)
+        # auto_confirm sets the `confirmed_at` field
+        if auto_confirm:
+            data["confirmed_at"] = datetime.now()
```
slint, line 202:
```suggestion
            data["confirmed_at"] = datetime.utcnow()
```

Based on [Flask-Security's `DATETIME_FACTORY`](https://github.com/inveniosoftware/flask-security-fork/blob/3001b0785d9ee769dae52484657e37e2733505db/flask_security/core.py#L124).

invenio_oauthclient/contrib/settings.py
slint: I think to full-proof this check, you would also need to check if "login without confirmation" is possible (maybe similar to [the first variable checks as done here](https://github.com/inveniosoftware/flask-security-fork/blob/3001b0785d9ee769dae52484657e37e2733505db/flask_security/confirmable.py#L61-L62)). That helps with the case where confirmation is required to login, and the configuration here would break it.

invenio_oauthclient/contrib/github.py
```diff
@@ -161,7 +167,22 @@ def _extract_email(gh):
...
+                username=user_info.login,
+                full_name=user_info.name,
+            ),
+        ),
+        external_id=str(user_info.id),
+        external_method="github",
+    )
+
+
+def account_info(remote, resp, info_serializer=default_info_serializer):
```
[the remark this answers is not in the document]
slint: I also have some concerns on this one, since e.g. the `default_info_serializer` in this case also accepts the extra `github_obj` parameter.

If it's for modularity I would say it's ok, but in that case I would make it explicit also that the "signature" accepts `**kwargs`.

slint (commented):
Looks great and also brings us closer to having a sane OAuth config :)

Some comments on correctness/edge-cases only.

invenio_oauthclient/utils.py
ntarocco: You are totally right, fixed it, thank you!

invenio_oauthclient/contrib/settings.py
ntarocco: In this case, the `CONFIRMABLE` will be true, so the `registration msg` must be sent right?
From the doc, `LOGIN_WITHOUT_CONFIRMATION` allows to login even before user's confirm, but the e-mail is sent.
```
Specifies if a user may login before confirming their email when the value of SECURITY_CONFIRMABLE is set to True
```

Honestly, I think we should remove this check completely, to avoid blocking any use case I cannot think ok... WDYT?

slint: Yeah, maybe it's best not to assume any of these (e.g. maybe someone wants to "manually confirm" users via backoffice), so let's remove :)

invenio_oauthclient/contrib/github.py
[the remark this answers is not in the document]
max-moser: oh, i think the "change" in my comment was ambiguous; i'm not proposing a change of the function
rather i'm just wondering if it's possible for users to override which `info_serializer` is being used by the function – if there is a place where users could pass a different argument, or if the function call is out of our hands (e.g. because it's called by a library?)

the splitting up of fetching and serializing logic makes total sense to me :+1:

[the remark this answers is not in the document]
slint: Ok, I thought I remembered this from somewhere, now I see where from :)

In that case, since it's just a convention within each contrib what parameters are being passed to their `info_serializer` I would leave it as is.

invenio_oauthclient/contrib/github.py
```diff
@@ -112,11 +113,10 @@ def __init__(
             access_token_url="https://github.com/login/oauth/access_token",
             authorize_url="https://github.com/login/oauth/authorize",
             precedence_mask=precedence_mask,
+            signup_options=signup_options,
         )
 
-    def get_handlers(self):
-        """Return GitHub auth handlers."""
-        return dict(
+        self.handlers = dict(
```
zzacharo, line 119: maybe we could add the `info_serializer` here as a config variable for the info handler?

invenio_oauthclient/contrib/github.py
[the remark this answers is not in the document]
ntarocco: Changes applied in a new commit, if you want to see the `diff`.

invenio_oauthclient/contrib/globus.py
```diff
@@ -79,8 +80,6 @@
 class GlobusOAuthSettingsHelper(OAuthSettingsHelper):
     """Default configuration for Globus OAuth provider."""
 
-    external_method = "globus"
```
ntarocco, line 82: I have removed all these hardcoded values, and replaced with `remote.name`, otherwise you cannot have 2 Globus apps configured (or others)... which might happen in weird cases! :)

zzacharo (approved):

Issue #296, opened 2023-02-23
update AAI endpoints/config


ppanero:
The current BETA instance of the OpenAIRE AAI OpenID Provider will be replaced by a new instance [1] based on Keycloak. All clients' configuration will be preserved along with their options (Client ID, Client Secret, Redirect URI(s), Scopes etc.), so there is no need to re-register your client(s). However, we need to reconfigure the OpenID Connect/OAuth 2.0 endpoints used by your Service, as described in our documentation [2].

Note that the Issuer Identifier for the new OpenAIRE AAI OpenID Provider has changed from "https://beta.aai.openaire.eu/oidc/" to "https://beta.aai.openaire.eu/auth/realms/openaire". Please read our migration guide [3] to see the full list of changes.

[1] https://beta.aai.openaire.eu/auth/realms/openaire/account
[2] https://openaire.github.io/login-docs/docs/developer/guide-for-sps#endpoints
[3] https://openaire.github.io/login-docs/docs/developer/guide-for-sps#client-migration-to-keycloak

ppanero:
**Waiting for answer**

At the moment we cannot upgrade due to errors with the User Info endpoint. The upgrade on the OpenAIRE side is to Keycloak, this means that on our side we also update the client.
The code changes can be found in this PR, but they should be moved to invenio-oauthclient once they are working and they are the following:

1- PR: https://github.com/zenodo/zenodo-rdm/pull/197
2- The REALM_URL and USER_INFO_URL also need to be defined as configuration variables. We do not have them there, but they are in the OpenShift config map:

  INVENIO_OAUTHCLIENT_OPENAIRE_AAI_VERIFY_EXP: 'True'
  INVENIO_OAUTHCLIENT_OPENAIRE_AAI_REALM_URL: 'https://beta.aai.openaire.eu/auth/realms/openaire’
  INVENIO_OAUTHCLIENT_OPENAIRE_AAI_USER_INFO_URL: >-
    https://beta.aai.openaire.eu/auth/realms/openaire/protocol/openid-connect/userinfo
  INVENIO_OAUTHCLIENT_OPENAIRE_AAI_VERIFY_AUD: 'False'

3- Note in the PR that we are adding two scopes (email and profile)

When logging in via AAI (Gmail, ORCID, or edugain+CERN) give the same error (Internal Error on our side after redirect), because the user_info does not have the `email` key.
[...]

PR #297, opened 2023-02-23
contrib: migrate openaire sandbox to keycloak

ppanero:
- closes https://github.com/inveniosoftware/invenio-oauthclient/issues/296

### DIFF
--- a/invenio_oauthclient/contrib/keycloak/handlers.py
+++ b/invenio_oauthclient/contrib/keycloak/handlers.py
@@ -45,11 +45,12 @@ from invenio_oauthclient.utils import oauth_link_external_id, oauth_unlink_exter
 from .helpers import get_user_info
 
 
-def info_serializer_handler(remote, resp, user_info, **kwargs):
+def info_serializer_handler(remote, resp, token_user_info, user_info=None, **kwargs):
     """Serialize the account info response object.
 
     :param remote: The remote application.
     :param resp: The response of the `authorized` endpoint.
+    :param token_user_info: The content of the authorization token response.
     :param user_info: The response of the `user info` endpoint.
     :returns: A dictionary with serialized user information.
     """
@@ -58,16 +59,25 @@ def info_serializer_handler(remote, resp, user_info, **kwargs):
     #
     # note: "external_id": `preferred_username` should also work,
     #       as it is seemingly not editable in Keycloak
+
+    user_info = user_info or {}  # prevent errors when accessing None.get(...)
+
+    email = token_user_info.get("email") or user_info["email"]
+    full_name = token_user_info.get("name") or user_info.get("name")
+    username = token_user_info.get("preferred_username") or user_info.get(
+        "preferred_username"
+    )
+
     return {
         "user": {
             "active": True,
-            "email": user_info["email"],
+            "email": email,
             "profile": {
-                "full_name": user_info.get("name"),
-                "username": user_info.get("preferred_username"),
+                "full_name": full_name,
+                "username": username,
             },
         },
-        "external_id": user_info["sub"],
+        "external_id": token_user_info["sub"],
         "external_method": remote.name,
     }
 
@@ -79,20 +89,20 @@ def info_handler(remote, resp):
     :param resp: The response of the `authorized` endpoint.
     :returns: A dictionary with the user information.
     """
-    user_info = get_user_info(remote, resp)
+    token_user_info, user_info = get_user_info(remote, resp)
 
     handlers = current_oauthclient.signup_handlers[remote.name]
     # `remote` param automatically injected via `make_handler` helper
-    return handlers["info_serializer"](resp, user_info)
+    return handlers["info_serializer"](resp, token_user_info, user_info)
 
 
 def setup_handler(remote, token, resp):
     """Perform additional setup after the user has been logged in."""
-    user_info = get_user_info(remote, resp)
+    token_user_info, _ = get_user_info(remote, resp)
 
     with db.session.begin_nested():
         # fetch the user's Keycloak ID and set it in extra_data
-        keycloak_id = user_info["sub"]
+        keycloak_id = token_user_info["sub"]
         token.remote_account.extra_data = {
             "keycloak_id": keycloak_id,
         }
--- a/invenio_oauthclient/contrib/keycloak/helpers.py
+++ b/invenio_oauthclient/contrib/keycloak/helpers.py
@@ -53,62 +53,59 @@ def get_public_key(remote, realm_url):
     return certs_resp["public_key"]
 
 
-def get_user_info(remote, resp_token, fallback_to_endpoint=True, options=dict()):
+def _get_user_info_from_token(remote, token, config_prefix):
+    """Get the user information from the JWT token."""
+    # try to parse the "id_token" part of Keycloak's (JWT) response
+    realm_url = current_app.config[f"{config_prefix}_REALM_URL"]
+    pubkey = _format_public_key(get_public_key(remote, realm_url))
+    alg = jwt.get_unverified_header(token)["alg"]
+
+    should_verify_aud = current_app.config.get(f"{config_prefix}_VERIFY_AUD", False)
+    expected_aud = current_app.config.get(f"{config_prefix}_AUD", None)
+
+    should_verify_expiration = current_app.config.get(
+        f"{config_prefix}_VERIFY_EXP", False
+    )
+
+    options = {
+        # check signature expiration
+        "verify_exp": should_verify_expiration,
+        # check the target audience
+        "verify_aud": should_verify_aud and (expected_aud is not None),
+    }
+
+    return jwt.decode(
+        token, key=pubkey, algorithms=[alg], audience=expected_aud, options=options
+    )
+
+
+def _get_user_info_from_endpoint(remote, config_prefix):
+    """Get the user info from the oauth server provider."""
+    url = current_app.config[f"{config_prefix}_USER_INFO_URL"]
+    return remote.get(url).data
+
+
+def get_user_info(remote, resp_token):
     """Get the user information from Keycloak.
 
     :param remote: The OAuthClient remote app
-    :param resp: The response from the 'token' endpoint; expected to be a dict
+    :param resp_token: The response from the 'token' endpoint; expected to be a dict
         and to contain a JWT 'id_token'
-    :param fallback_to_endpoint: Whether or not to fall back to the 'userinfo'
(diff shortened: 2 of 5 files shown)


ppanero:
blocked by https://zenodo-sentry.web.cern.ch/sentry/zenodo-rdm/issues/19978243/?query=is%3Aunresolved needs debugging

invenio_oauthclient/contrib/keycloak/helpers.py
```diff
@@ -53,7 +53,45 @@ def get_public_key(remote, realm_url):
     return certs_resp["public_key"]
 
 
-def get_user_info(remote, resp_token, fallback_to_endpoint=True, options=dict()):
```
ppanero, line 56: `fallback_to_endpoint=True, options=dict()` were actually unreachable, since both the place where this function is called and the corresponding handler does not hage `*args, **kwargs` (and therefore does not forward them neither).

invenio_oauthclient/contrib/keycloak/helpers.py
```diff
@@ -64,51 +102,13 @@ def get_user_info(remote, resp_token, fallback_to_endpoint=True, options=dict())
...
-            options.update({"verify_exp": True})
-        else:
-            options.update({"verify_exp": False})
-
-        user_info = jwt.decode(
-            token, key=pubkey, algorithms=[alg], audience=expected_aud, options=options
-        )
-
-    except Exception as e:
-        if not fallback_to_endpoint:
```
ppanero, line 101: Fallback is always desired, if the fallback fails it will be raised in any case.

invenio_oauthclient/contrib/keycloak/helpers.py
```diff
@@ -64,51 +102,13 @@ def get_user_info(remote, resp_token, fallback_to_endpoint=True, options=dict())
...
-
-        # as a fallback, we can still contact Keycloak's userinfo endpoint
-        # `remote.get(...)` automatically includes OAuth2 tokens in the header
-        # and the response's `data` field is a dict
-        url = current_app.config[f"{config_prefix}_USER_INFO_URL"]
-        user_info = remote.get(url).data
-
-    return user_info
+    use_token = current_app.config[f"{config_prefix}_USER_INFO_FROM_TOKEN"]
+    if use_token:
```
ppanero, line 106: Optionally allowing trying to find the information in the initial JWT token. For OpenAIRE we do not want this since it lacks some information (configured per instance).

invenio_oauthclient/contrib/keycloak/helpers.py
```diff
@@ -53,7 +53,45 @@ def get_public_key(remote, realm_url):
...
+    return jwt.decode(
+        token, key=pubkey, algorithms=[alg], audience=expected_aud, options=options
+    )
+
+def _get_user_info_from_endpoint(remote, config_prefix):
+    """Get the user info from the oauth server provider."""
+    url = current_app.config[f"{config_prefix}_USER_INFO_URL"]
+    return remote.get(url).data
+
+def get_user_info(remote, resp_token):
```
max-moser, line 94: minor: the docstring should be updated accordingly

(looks like there never was a `resp` argument... i think it references `resp_token` :sweat_smile: )

invenio_oauthclient/contrib/keycloak/helpers.py
```diff
@@ -53,7 +53,45 @@ def get_public_key(remote, realm_url):
...
 
-def get_user_info(remote, resp_token, fallback_to_endpoint=True, options=dict()):
+def _get_user_info_from_token(remote, token, config_prefix):
+    """Get the user information from the JWT token."""
+    # try to parse the "id_token" part of Keycloak's (JWT) response
+    realm_url = current_app.config[f"{config_prefix}_REALM_URL"]
+    pubkey = _format_public_key(get_public_key(remote, realm_url))
+    alg = jwt.get_unverified_header(token)["alg"]
+
+    if not isinstance(options, dict):
```
max-moser, line 63: won't this crash because `options` isn't defined here yet?
just setting `options = {}` should be fine

invenio_oauthclient/contrib/keycloak/helpers.py
max-moser: i can't remember why i made that optional in the first place but i guess it was intended as a switch for either:
* saving network requests
* reducing communication with external services for security

given that the URL to be called is defined in the app config (and thus controlled by the admin) however, that should be okay as long as the communication uses valid SSL certificates

invenio_oauthclient/contrib/keycloak/helpers.py
max-moser: these arguments are passed in the tests: https://github.com/inveniosoftware/invenio-oauthclient/blob/master/tests/test_contrib_keycloak.py#L381

though the `options` there look like they could be replaced via tweaking `app.config` a bit
and the fallback endpoint shouldn't be called in the test anyway (`fallback_to_endpoint=False`)... could be mocked with `httpretty` to return a failure to make sure it's not called

invenio_oauthclient/contrib/keycloak/helpers.py
```diff
@@ -64,51 +102,13 @@ def get_user_info(remote, resp_token, fallback_to_endpoint=True, options=dict())
...
-        url = current_app.config[f"{config_prefix}_USER_INFO_URL"]
-        user_info = remote.get(url).data
-
-    return user_info
+    use_token = current_app.config[f"{config_prefix}_USER_INFO_FROM_TOKEN"]
+    if use_token:
+        try:
+            return _get_user_info_from_token(remote, resp_token["id_token"], config_prefix)
+        except Exception as e:
+            pass
```
max-moser, line 110: minor: should we maybe log the exception instead of just swallowing it?

max-moser (approved):
the general refactor is looking good to me, just a few minor remarks

ppanero:
tests are failing on flask babelex

invenio_oauthclient/contrib/keycloak/helpers.py
ntarocco: I would propose an alternative solution:
- Always fetch user info from remote. Are we sure we want a try/except Exception here? The keycloack server should always have the user endpoint, and if it failed for any reason, the user should try again.
- Then, take that response, and pass it to the user info serializer: in your instance, you can implement the user serializer and then decide that you don't want to trust data coming from remote or you want to complement them.

invenio_oauthclient/contrib/openaire_aai.py
```diff
@@ -160,8 +161,13 @@ def get_rest_handlers(self):
 """OpenAIRE Remote REST Application."""
 
 # Sandbox
-_openaire_aai_sandbox_app = OpenAIREAuthSettingsHelper(
-    base_url="https://openaire-dev.aai-dev.grnet.gr",
+_openaire_aai_sandbox_app = KeycloakSettingsHelper(
```
ntarocco, line 164: This will probably require a major release. Alternative is to create a new contrib.

ppanero: OpenAIRE has decomissioned `https://openaire-dev.aai-dev.grnet.gr` (the old URL for the sandbox). So nobody can use it and if in use it is failing. I think in this circumstances could even be a "patch" 😂, I would still do it as minor. WDYT?

max-moser:
It's been quite a while since I first wrote the Keycloak integration, and it _was_ my first real contribution to InvenioRDM...
So I can't really tell with certainty anymore what the precise thoughts were behind it all :sweat_smile: 

But IIRC, I just tried to save a few unnecessary network requests and reduce the attack surface.
The user info endpoint probably was just added as a kind of "might as well just try that" in case something goes wrong with the initially passed token (e.g. if the encryption was invalid, ...).

ntarocco (commented):
As discussed:
- always get info from token
- optionally, get info from remote, via config var
- the `info_serializer_handler` hook should get both, token and remote (can be None) info, and decide how to merge them

invenio_oauthclient/contrib/keycloak/handlers.py
```diff
@@ -45,7 +45,7 @@
 from .helpers import get_user_info
 
 
-def info_serializer_handler(remote, resp, user_info, **kwargs):
+def info_serializer_handler(remote, resp, token_user_info, user_info, **kwargs):
```
ppanero, line 48: breaking change: the order does not change since previously the user info was from the token, however, the name of the paramter changes.

invenio_oauthclient/contrib/keycloak/helpers.py
```diff
@@ -53,62 +53,57 @@ def get_public_key(remote, realm_url):
...
+    url = current_app.config[f"{config_prefix}_USER_INFO_URL"]
+    return remote.get(url).data
+
+
+def get_user_info(remote, resp_token):
     """Get the user information from Keycloak.
 
     :param remote: The OAuthClient remote app
-    :param resp: The response from the 'token' endpoint; expected to be a dict
+    :param resp_token: The response from the 'token' endpoint; expected to be a dict
```
ntarocco, line 92: can you please add a docstring about the `return` types?

invenio_oauthclient/contrib/keycloak/handlers.py
ntarocco, line 48: Very minor:
```suggestion
def info_serializer_handler(remote, resp, token_user_info=None, user_info=None, **kwargs):
```
It looks like, from the code, that they could be both `None` (in case of errors and/or the remote fetch is disabled).
If that's the case, I think this should be reflected in the code and implementation of this method.

Ideally, I would have `token_user_info` always set, with default `{}`, and `user_info` to be `{}` or `None`... maybe `None` would be better when disabled, it means that it was not called (while `{}` means that it was called, but no values returned).

invenio_oauthclient/contrib/keycloak/handlers.py
```diff
@@ -58,16 +59,23 @@ def info_serializer_handler(remote, resp, user_info, **kwargs):
     #
     # note: "external_id": `preferred_username` should also work,
     #       as it is seemingly not editable in Keycloak
+
```
ntarocco, line 62: As per chat:
```suggestion
user_info = user_info or {...}
```

ntarocco (approved):

invenio_oauthclient/contrib/openaire_aai.py
kpsherva: IMO still major, since we want to follow semver, I think it is regardless external condidtions :)

PR #302, opened 2023-04-11
base: add group handler 

### LINKED ISSUE #245: Groups membership support

TLGINO:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2186
* NEEDS https://github.com/inveniosoftware/invenio-accounts/pull/439

### DIFF
--- a/invenio_oauthclient/handlers/base.py
+++ b/invenio_oauthclient/handlers/base.py
@@ -10,8 +10,9 @@
 
 from flask import current_app, session
 from flask_login import current_user
+from invenio_accounts.models import Role
+from invenio_accounts.proxies import current_datastore
 from invenio_db import db
-from pkg_resources import require
 
 from ..errors import (
     OAuthClientAlreadyAuthorized,
@@ -46,6 +47,56 @@ from .utils import (
 )
 
 
+def _role_needs_update(role_obj, new_role_dict):
+    """Checks if role needs to be updated."""
+    if role_obj.name != new_role_dict.get(
+        "name"
+    ) or role_obj.description != new_role_dict.get("description"):
+        return True
+    return False
+
+
+def create_or_update_groups(account_groups):
+    """Creates the roles based on the groups provided."""
+    roles_id = []
+    for group in account_groups:
+        existing_role = current_datastore.find_role_by_id(group["id"])
+        if existing_role and existing_role.is_managed:
+            current_app.logger.exception(
+                f'Error while syncing roles: A managed role with id: ${group["id"]} already exists'
+            )
+            continue
+        existing_role_by_name = current_datastore.find_role(group["name"])
+        if existing_role_by_name and existing_role_by_name.is_managed:
+            current_app.logger.exception(
+                f'Error while syncing roles: A managed role with name: ${group["name"]} already exists'
+            )
+            continue
+        if not existing_role:
+            role = current_datastore.create_role(
+                id=group["id"],
+                name=group.get("name"),
+                description=group.get("description"),
+                is_managed=False,
+            )
+            roles_id.append(role.id)
+        elif existing_role and _role_needs_update(existing_role, group):
+            role_to_update = Role(
+                id=group["id"],
+                name=group.get("name"),
+                description=group.get("description"),
+                is_managed=False,
+            )
+            role = current_datastore.update_role(role_to_update)
+            roles_id.append(role.id)
+        else:
+            roles_id.append(existing_role.id)
+
+    current_datastore.commit()
+
+    return roles_id
+
+
 #
 # Handlers
 #
@@ -53,7 +104,7 @@ def base_authorized_signup_handler(resp, remote, *args, **kwargs):
     """Handle sign-in/up functionality.
 
     :param remote: The remote application.
-    :param resp: The response.
+    :param resp: The response of the `authorized` endpoint.
     :returns: Redirect response.
     """
     # Remove any previously stored auto register session key
@@ -65,11 +116,12 @@ def base_authorized_signup_handler(resp, remote, *args, **kwargs):
     # current_user.is_authenticated().
     token = response_token_setter(remote, resp)
     handlers = current_oauthclient.signup_handlers[remote.name]
-
-    # Sign-in/up user
-    # ---------------
+    # Needed for tests
     if not current_user.is_authenticated:
+        # Sign-in/up user
+        # ---------------
         account_info = handlers["info"](resp)
+        assert "external_id" in account_info
         account_info_received.send(
             remote, token=token, response=resp, account_info=account_info
         )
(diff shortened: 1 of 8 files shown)


invenio_oauthclient/contrib/keycloak/handlers.py
```diff
@@ -96,6 +96,31 @@ def info_handler(remote, resp):
     return handlers["info_serializer"](resp, token_user_info, user_info)
 
 
+def groups_serializer_handler(remote, resp, user, groups, **kwargs):
```
jrcastro2, line 99: Note: This has to be removed before merging!

invenio_oauthclient/contrib/keycloak/settings.py
```diff
@@ -73,6 +73,8 @@ def __init__(
             signup_handler=dict(
                 info="invenio_oauthclient.contrib.keycloak.handlers:info_handler",
                 info_serializer="invenio_oauthclient.contrib.keycloak.handlers:info_serializer_handler",
+                groups="invenio_oauthclient.contrib.keycloak.handlers:groups_handler",
```
jrcastro2, line 76: Note: This has to be removed before merging!

invenio_oauthclient/contrib/keycloak/settings.py
```diff
@@ -84,6 +86,8 @@ def __init__(
...
```
jrcastro2, line 89: Note: This has to be removed before merging!

TLGINO:
List of dependencies:

image


Tests passing localy:


image


Disclaimer: Alembic tests marked as skipped

invenio_oauthclient/handlers/base.py
```diff
@@ -103,7 +114,48 @@ def base_authorized_signup_handler(resp, remote, *args, **kwargs):
                 db.session.commit()
                 raise OAuthClientMustRedirectSignup()
 
-        # Authenticate user
+        account_groups = handlers["groups"](resp, user)
+        if account_groups:
```
ntarocco, line 118: can we extract in a different func maybe?

invenio_oauthclient/handlers/base.py
```diff
@@ -103,7 +114,48 @@ def base_authorized_signup_handler(resp, remote, *args, **kwargs):
...
+            roles_id = []
+            for group in account_groups:
+                existing_role = current_datastore.find_role_by_id(group["id"])
+                if existing_role and existing_role.is_managed:
+                    current_app.logger.exception(
+                        f'Error while syncing roles: A managed role with id: ${group["id"]} already exists'
+                    )
+                    continue
+                existing_role_by_name = current_datastore.find_role(group["name"])
```
ntarocco, line 127: question: why do we check if the name is existing? It does make much sense, but maybe it could be that 2 roles have different IDs, but same name? Or not?

jrcastro2: No, name has to be unique (here the [model](https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/models.py#L69)), therefore we check if there is an already existing managed role to avoid conflicts. I adde those checks because if there is an error it would rollback all the changes. An alternative would be commit one by one.

tests/test_handlers_rest.py
```diff
@@ -82,9 +82,9 @@ def test_unauthorized_signup(remote, app_rest, models_fixture):
     example_account_info = {
         "user": {
             "email": existing_email,
-            "external_id": "1234",
-            "external_method": "test_method",
-        }
+        },
+        "external_id": "1234",
```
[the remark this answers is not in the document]
jrcastro2: This is the consequence of this [change](https://github.com/inveniosoftware/invenio-oauthclient/pull/302/files#diff-7a8682eb341edb9f0e2208e36de3df79d569df2dd0bebedfaca72994d498cabbR83). Copied from your [initial PR](https://github.com/inveniosoftware/invenio-oauthclient/pull/290/files#diff-7a8682eb341edb9f0e2208e36de3df79d569df2dd0bebedfaca72994d498cabbR73).
If the `assert` is correct then this change is correct as well.

invenio_oauthclient/handlers/base.py
```diff
@@ -103,7 +114,48 @@ def base_authorized_signup_handler(resp, remote, *args, **kwargs):
                 db.session.commit()
                 raise OAuthClientMustRedirectSignup()
 
-        # Authenticate user
+        account_groups = handlers["groups"](resp, user)
+        if account_groups:
+            roles_id = []
+            for group in account_groups:
```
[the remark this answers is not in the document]
jrcastro2: More or less this is what's it's happening, the Role objects are added to the session and committed all at once at the very end. And the records will be indexed as well all at once (this is being addressed by Martin in [this PR](https://github.com/inveniosoftware/invenio-users-resources/issues/76)). However, I agree with you, I will try to improve this and remove all these checks

invenio_oauthclient/handlers/base.py
```diff
@@ -46,14 +47,63 @@
...
+    if role_obj.name != new_role_dict.get("name") or role_obj.description != new_role_dict.get("description"):
+        return True
+    return False
+
+
+def _create_or_update_groups(account_groups):
+    """Creates the roles based on the groups provided."""
+    roles_id = []
+    for group in account_groups:
+        existing_role = current_datastore.find_role_by_id(group["id"])
```
ntarocco, line 61: I thought I have added a comment here, but I don't see it.
I think that we should avoid this 2 checks, for performance reasons. This happens during the HTTP request of the login: for users with high number of groups, it adds extra delays.

Can we try to insert and catch the failure instead?

invenio_oauthclient/handlers/base.py
```diff
@@ -46,14 +47,63 @@
...
+            )
+            role = current_datastore.update_role(role_to_update)
+            roles_id.append(role.id)
+        else:
+            roles_id.append(existing_role.id)
+
+    current_datastore.commit()
+
+    # We set the unmanaged groups in the session because they are not linked to the user in the DB
+    session["_unmanaged_groups"] = roles_id
```
ntarocco, line 96: I am a bit worried by this func, let's chat IRL.

invenio_oauthclient/handlers/base.py
```diff
@@ -46,14 +47,63 @@
...
+                id=group["id"],
+                name=group.get("name"),
+                description=group.get("description"),
+                is_managed=False,
+            )
+            roles_id.append(role.id)
+        elif existing_role and _role_needs_update(existing_role, group):
+            role_to_update = Role(
+                id=group["id"],
+                name=group.get("name"),
```
ntarocco, line 84: As discussed, just double check that it can be `null`.

invenio_oauthclient/handlers/base.py
```diff
@@ -46,14 +47,63 @@
...
+    """Creates the roles based on the groups provided."""
+    roles_id = []
+    for group in account_groups:
+        existing_role = current_datastore.find_role_by_id(group["id"])
+        if existing_role and existing_role.is_managed:
+            current_app.logger.exception(
+                f'Error while syncing roles: A managed role with id: ${group["id"]} already exists'
+            )
+            continue
+        existing_role_by_name = current_datastore.find_role(group["name"])
```
ntarocco, line 67: can we check if an `index` can be added here when the task to import groups in CDS is done?

invenio_oauthclient/handlers/base.py
jrcastro2: Confirmed, name can be null 
image

tests/test_app.py
```diff
@@ -123,6 +123,9 @@ def teardown():
         assert len(tables) == 2
 
 
+@pytest.mark.skip(
```
ppanero, line 126: Remove after merging access/accounts

ppanero (commented):
I see a `groups_handler` test, however I'm not sure how feasible is to implement that. Maybe checking the feasibility of implementing a test to check it is called? (e.g. using `assert_called`)

PR #303, opened 2023-05-13
fix: before_first_request deprecation

utnapischtim:
invenio-theme, flask-menu, invenio-accounts, invenio-userprofiles


invenio_oauthclient/views/client.py
```diff
@@ -39,28 +39,6 @@
 )
 
 
-@blueprint.record_once
```
slint, line 42: One thing I just realized, in this case this function using `@blueprint.record_once` would only run for the UI app and not the API app (i.e. the `/api` mount). Some of these cases are intentional (like here for example, we're setting template config variables). For other cases it might be though something that was overlooked or a even mistake.

This might hint that we need a UI and API variant of the "finalize_app" entrypoint? Similar to how we have `invenio_base.blueprints` and `invenio_base.api_blueprints`.

invenio_oauthclient/finalize_app.py
```diff
@@ -0,0 +1,83 @@
...
+        app.config["OAUTHCLIENT_LOGIN_USER_TEMPLATE_PARENT"] = template
+        app.config[template_key] = app.config.get(
+            "OAUTHCLIENT_LOGIN_USER_TEMPLATE",
+            "invenio_oauthclient/login_user.html",
+        )
+
+
+def init_index_menu(app):
+    """Init index menu."""
+    item = current_menu.submenu("settings.oauthclient")
```
slint, line 43: Isn't the use of `current_menu` here wrong? Since we're not inside the application context at this point AFAIK.

Wouldn't the alternative be to either call `app.extensions['menu'].submenu(...)` or have the un-initialized `Menu` extension to import from invenio-theme?

slint (commented):
LGTM, noted a quick potential change to the entrypoints for having a UI/API variant.

slint (commented):
LGTM, noted a quick potential change to the entrypoints for having a UI/API variant.

invenio_oauthclient/finalize_app.py
utnapischtim: hm, but the code as it is is working, but it would be more understandable if i we change to `app.extensions['menu'].submenu(...)`

invenio_oauthclient/finalize_app.py
```diff
@@ -0,0 +1,83 @@
...
+        ),
+        visible_when=lambda: bool(app.config.get("OAUTHCLIENT_REMOTE_APPS"))
+        is not False,
+    )
+
+
+def init_index_breadcrumb():
+    """Init index breadcrumb."""
+    item = current_menu.submenu("breadcrumbs.settings.oauthclient")
+    item.register("invenio_oauthclient_settings.index", _("Linked accounts"))
```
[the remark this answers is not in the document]
utnapischtim: maybe i will write they in a dict and do it with for loop. it is sometimes a little bit difficult to read, if everything is written out.

invenio_oauthclient/views/client.py
utnapischtim: should i have first a look, if there are already use-cases for `invenio_base.api_finalize_app`?

slint: I think it's good to have it and be conscious about its usage, so yes I mean it's meant to happen at some point that we need something running on the API app only.

invenio_oauthclient/finalize_app.py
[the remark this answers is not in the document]
slint: I'm fine with anything as long as it reads clean :)

It's more about the fact that so far usually wherever I see a `register_menu`, there's also `register_breadcrumb`, so it kind of makes sense to do these together.

invenio_oauthclient/views/client.py
utnapischtim: the [commit](https://github.com/inveniosoftware/invenio-app/blob/5db9f3349f4593a108d67fb1be07370f310ef419/invenio_app/factory.py#L91-L100) in invenio-app which introduce the `invenio_base.finalize_app` entrypoint is only applied to `create_ui` and `create_app` and not `create_api`. Should the `invenio_base.api_finalize_app` entrypoint then added to `create_api`? 

the `/api` endpoint is mentioned in `create_app`. does this mean that `invenio_base.api_finalize_app` should be added to `create_app` instead of `invenio_base.finalize_app`? As i read that code, that would be false, so it should remain as it is with `invenio_base.finalize_app`.

invenio_oauthclient/finalize_app.py
utnapischtim: both variants `current_menu` and `app.extensions["menu"].submenu(...)` are working as expected. i tested it by changing the text and checking the value on the page (not the best way, but for now i think enough). further i thought that `current_menu` has to work, because otherwise the application would not start.

PR #304, opened 2023-05-26
client: add logout blueprint

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/16

### DIFF
--- a/invenio_oauthclient/handlers/base.py
+++ b/invenio_oauthclient/handlers/base.py
@@ -109,7 +109,9 @@ def base_authorized_signup_handler(resp, remote, *args, **kwargs):
     """
     # Remove any previously stored auto register session key
     session.pop(token_session_key(remote.name) + "_autoregister", None)
-
+    # We set the remote in the session to be aware of which one is being used and, on log out redirect to
+    # the correct URL set in the OAUTHCLIENT_REMOTE_APPS for each remote
+    session["OAUTHCLIENT_SESSION_REMOTE_NAME"] = remote.name
     # Store token in session
     # ----------------------
     # Set token in session - token object only returned if
--- a/invenio_oauthclient/views/client.py
+++ b/invenio_oauthclient/views/client.py
@@ -8,7 +8,7 @@
 
 """Client blueprint used to handle OAuth callbacks."""
 
-from flask import Blueprint, abort, current_app, redirect, request, url_for
+from flask import Blueprint, abort, current_app, redirect, request, session, url_for
 from flask_oauthlib.client import OAuthException
 from invenio_accounts.views import login as base_login
 from invenio_db import db
@@ -271,3 +271,20 @@ def rest_disconnect(remote_app):
         return _disconnect(remote_app)
     except OAuthRemoteNotFound:
         abort(404)
+
+
+@blueprint.route("/logout")
+def post_logout():
+    """Client logout view.
+
+    This URL should be called by setting `SECURITY_POST_LOGOUT_VIEW = /oauth/logout`
+    """
+    remote_name = session.pop("OAUTHCLIENT_SESSION_REMOTE_NAME", None)
+    if remote_name:
+        logout_url = current_app.config["OAUTHCLIENT_REMOTE_APPS"][remote_name].get(
+            "logout_url"
+        )
+        if logout_url:
+            return redirect(logout_url, code=302)
+
+    return redirect("/")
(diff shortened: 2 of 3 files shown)


invenio_oauthclient/views/client.py
```diff
@@ -271,3 +271,19 @@ def rest_disconnect(remote_app):
         return _disconnect(remote_app)
     except OAuthRemoteNotFound:
         abort(404)
+
+
+@blueprint.route("/logout")
+def logout():
```
ntarocco, line 277: can we make the implementation of this endpoint configurable, in case someone would like to implement a different behaviour? (e.g. add extra params?)
How can I, for example, add a redirect back to a specific URL (which might not be the homepage)?

invenio_oauthclient/handlers/base.py
```diff
@@ -58,7 +58,9 @@ def base_authorized_signup_handler(resp, remote, *args, **kwargs):
     """
     # Remove any previously stored auto register session key
     session.pop(token_session_key(remote.name) + "_autoregister", None)
-
+    # We set the remote in the session to be aware of which one is being used and, on log out redirect to
+    # the correct URL set in the OAUTHCLIENT_REMOTE_APPS for each remote
+    session["remote_name"] = remote.name
```
ntarocco, line 63: does it make sense to namespace this, with a prefix?
See for example `current_app.config["OAUTHCLIENT_SESSION_KEY_PREFIX"]`, I don't know if it makes much sense.

invenio_oauthclient/views/client.py
```diff
@@ -271,3 +271,19 @@ def rest_disconnect(remote_app):
...
+    """Client logout view.
+
+    This URL should be called by setting `SECURITY_POST_LOGOUT_VIEW = /oauth/logout`
+    """
+    remote_name = session.pop("remote_name")
```
ntarocco, line 282: I am wondering if it could be possible that this is non-existing... do you think it might makes sense to check that?
```
remote_name = session.pop("remote_name", None)
if remote_name:
    ...
```

invenio_oauthclient/views/client.py
kpsherva, line 277: could be `post_logout`, otherwise naming a bit confusing

invenio_oauthclient/handlers/base.py
slint: Agree on this one... Also, what happens e.g. when you connect multiple OAuth providers to your account? Would it be only the last one that ends up in the session variable? Would it make sense to keep a list/set of remotes like so:

```python
session.setdefault("remote_names", set())  # not sure if set is supported
session["remote_name"].add(remote.name)
```

Then you could in logout check which of the remotes has a `logout_url` specified and again make an arbitrary choice, e.g. the first one...

This sounds a bit overcomplicated, and I'm not sure if it makes perfect sense though.

invenio_oauthclient/views/client.py
jrcastro2: This endpoint is only used if the variable `SECURITY_POST_LOGOUT_VIEW=/oauth/logout` is set, meaning that if you want to configure to redirect to another endpoint you just need to register the endpoint and override the `SECURITY_POST_LOGOUT_VIEW` with the new value. To add the redirect back to a specific URL I think it's each remote responsability to do it, for instance in [this case (cds-rdm)](https://github.com/CERNDocumentServer/cds-rdm/pull/66/files#diff-35ea7d0c9f0d3771b33ec0ac671404047bdff444801fcf084c39437c21107ebfR213) you pass the `redirect_uri` pointing to the home page

invenio_oauthclient/handlers/base.py
jrcastro2: Prefixed as requested.
I am not sure if multiple accounts connected affects this, the remote_name is set on loging and although your account might be connected to multiple OAuth providers you always log in with only one of the, or am I wrong?

However, if I am wrong, looks to me that could be done in another issue as an improvement (?)

invenio_oauthclient/views/client.py
```diff
@@ -271,3 +271,19 @@ def rest_disconnect(remote_app):
         return _disconnect(remote_app)
     except OAuthRemoteNotFound:
         abort(404)
+
+
+@blueprint.route("/logout")
```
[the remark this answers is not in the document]
jrcastro2: If we decide to got for this I think we should split between `logout_url´ and `rest_logout_url` because I guess while using the rest you would like to redirect to rest endpoint as well ?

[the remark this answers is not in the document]
kpsherva: lets do as an improvement and ship this one

kpsherva (approved):

PR #326, opened 2024-04-11
global: explicitly set and use hide attribute for config

rekt-hard:
:heart: Thank you for your contribution!

### Description

A `hide` attribute is accessed at some places within the existing code already. These changes will explicitly set this attribute and use it in other places in the code as well. This will allow to hide login with an external account, while at the same time, allow to link an external account in the profile view.

Per default, `hide` will evaluate to `False` in order to be compliant with the current logic.

Example config:
```py

def anon_user():
    return not current_user.is_authenticated

helper = k.KeycloakSettingsHelper(
    title="Local KeyCloak",
    description="My organization authentication provider",
    base_url="http://localhost:8087",
    realm="damap",
    # hide_when=anon_user
)

OAUTHCLIENT_REMOTE_APPS = dict(
    orcid={
        **orcid.REMOTE_APP,
        "hide": LocalProxy(lambda: anon_user()),
    },
    keycloak={
        **helper.remote_app,
    },
)
```

Login only via local keycloak:
image

Linking of accounts with ORCID also possible:
image

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [ ] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [ ] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [ ] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/) (for relevant code).
- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [ ] I've NOT included third-party code (copy/pasted source code or new dependencies).

**Third-party code**

If you've added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/invenio_oauthclient/contrib/settings.py
+++ b/invenio_oauthclient/contrib/settings.py
@@ -8,6 +8,7 @@
 """Common OAuth configuration helper."""
 
 from werkzeug import cached_property
+from werkzeug.local import LocalProxy
 
 
 class OAuthSettingsHelper:
@@ -31,6 +32,7 @@ class OAuthSettingsHelper:
         precedence_mask=None,
         signup_options=None,
         logout_url=None,
+        hide_when=False,
         **kwargs,
     ):
         """The constructor."""
@@ -51,6 +53,7 @@ class OAuthSettingsHelper:
         signup_options = signup_options or {}
         signup_options.setdefault("auto_confirm", True)
         signup_options.setdefault("send_register_msg", False)
+        self.hide_when = hide_when if hide_when is not None else False
 
         self.base_app = dict(
             title=title,
@@ -59,6 +62,9 @@ class OAuthSettingsHelper:
             precedence_mask=precedence_mask,
             signup_options=signup_options,
             logout_url=logout_url,
+            hide=LocalProxy(
+                lambda: self.hide_when() if callable(hide_when) else self.hide_when
+            ),
             params=dict(
                 base_url=self.base_url,
                 request_token_params=request_token_params,
--- a/invenio_oauthclient/views/client.py
+++ b/invenio_oauthclient/views/client.py
@@ -60,12 +60,19 @@ def auto_redirect_login(*args, **kwargs):
     )
     would_redirect = auto_redirect_enabled and not local_login_enabled
     remote_apps = list(current_oauthclient.oauth.remote_apps)
+    remote_app_configs = current_app.config["OAUTHCLIENT_REMOTE_APPS"]
 
-    if would_redirect and len(remote_apps) == 1:
+    visible_remote_apps = [
+        remote_app
+        for remote_app in remote_apps
+        if remote_app_configs.get(remote_app, {}).get("hide", False)
+    ]
+
+    if would_redirect and len(visible_remote_apps) == 1:
         redirect_args = {
             # if local login is disabled and we only have one OAuth2 remote app
             # configured, we forward directly to that
-            "remote_app": remote_apps[0],
+            "remote_app": visible_remote_apps[0],
         }
         next_url = request.args.get("next")
         if next_url:
@@ -109,6 +116,13 @@ def _login(remote_app, authorized_view_name):
 @blueprint.route("/login/<remote_app>/")
 def login(remote_app):
     """Send user to remote application for authentication."""
+    if (
+        current_app.config["OAUTHCLIENT_REMOTE_APPS"]
+        .get(remote_app, {})
+        .get("hide", False)
+    ):
+        abort(404)
+
     try:
         return _login(remote_app, ".authorized")
     except OAuthRemoteNotFound:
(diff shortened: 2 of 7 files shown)


invenio_oauthclient/contrib/settings.py
```diff
@@ -8,6 +8,12 @@
 """Common OAuth configuration helper."""
 
 from werkzeug import cached_property
+from werkzeug.local import LocalProxy
+
+
+def CONDITION_FALSE(*args, **kwargs):
```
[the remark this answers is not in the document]
rekt-hard: This is used as a default value for the `visible_when` attribute here: https://github.com/inveniosoftware/invenio-oauthclient/pull/326/files#diff-335ed662258b6b6bbfbca87e02ab6ae65277accaa70462f6069b3915de4b39caR61

Same usage can be found in flask menu:
https://github.com/inveniosoftware/flask-menu/blob/9fce2f84380e9699131421e41b4428d05d39618c/flask_menu/menu.py#L24
https://github.com/inveniosoftware/flask-menu/blob/9fce2f84380e9699131421e41b4428d05d39618c/flask_menu/menu.py#L182-L184

If preferred, we can refactor to it to be lower case snake case. Can also try to import it from there :)

invenio_oauthclient/contrib/settings.py
```diff
@@ -59,6 +67,7 @@ def __init__(
             precedence_mask=precedence_mask,
             signup_options=signup_options,
             logout_url=logout_url,
+            hide=LocalProxy(lambda: self.hide_when()),
```
max-moser, line 70:
```suggestion
            hide=LocalProxy(lambda: self.hide_when() if callable(self.hide_when) else self.hide_when),
```
Supporting simple values next to callables could simplify the configuration a bit and drop the requirement of the `CONDITION_FALSE` import/definition above.
It would, of course, require an extra line of documentation.

invenio_oauthclient/contrib/settings.py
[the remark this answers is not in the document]
rekt-hard: Removed due to support for simple values

max-moser (approved):

max-moser:
@rekt-hard suggested disabling the `/oauth/<remote-app>/` endpoint if the `hide` config for `<remote-app>` evaluates to `True`.
Otherwise, users can still use that OP to authenticate, even if it's hidden.
I'll merge this PR once that functionality is added.

For dynamic cases (e.g. only allow a given OP for linking but not registering), see the example configuration.

max-moser (approved):

invenio_oauthclient/views/client.py
```diff
@@ -60,8 +60,13 @@ def auto_redirect_login(*args, **kwargs):
...
     would_redirect = auto_redirect_enabled and not local_login_enabled
     remote_apps = list(current_oauthclient.oauth.remote_apps)
+    remote_app_configs = current_app.config["OAUTHCLIENT_REMOTE_APPS"]
 
-    if would_redirect and len(remote_apps) == 1:
+    if (
+        would_redirect
+        and len(remote_apps) == 1
+        and not remote_app_configs[0].get("hide", False)
+    ):
```
max-moser, line 69: AFAIK, the `hide` flag wasn't really exposed before this PR so i doubt that this makes a difference to anybody (i expect all configured `REMOTE_APPS` to be visible).

would it make sense here to filter the `remote_apps` and `remote_app_configs` to only the visible ones here before this check?
otherwise, if you have one hidden and one visible remote app configured, this logic would never trigger (even though i would expect it to?)

rekt-hard: Thank you, good catch. Adapted it to perform the check on visible remote apps only

PR #328, opened 2024-05-14
OAuth2 Token refresh implemented

### LINKED ISSUE #68: oauthclient: handle access token expiration
@lnielsen commented on [Mon Aug 18 2014](https://github.com/inveniosoftware/invenio/issues/2114)

- [ ] Store refresh token if provided by provider.
- [ ] Store expire time if provided by provider.
- [ ] Add method to refresh access token in ``RemoteToken`` model.

mesemus:
**Important**: we are aiming to release this at the same time as https://github.com/inveniosoftware/invenio-oauthclient/pull/360. When we merge, we need to align the DB migrations, but still keep them as 2 separate migrations so that one can be run manually. This migration does not need to be run manually in any scenario.

### Description

Closes #68.

Added fields on `RemoteToken` model:
* `refresh_token`
* `expires_at`

Added properties on `RemoteToken`:
* `is_expired`

New methods on `RemoteToken`:
* `refresh_access_token`

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/invenio_oauthclient/alembic/7def990b852e_add_expires_at_and_refresh_token_to_.py
+++ b/invenio_oauthclient/alembic/7def990b852e_add_expires_at_and_refresh_token_to_.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2016-2025 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Add expires and refresh_token to remote token."""
+
+import sqlalchemy as sa
+import sqlalchemy_utils
+from alembic import op
+from sqlalchemy.dialects import mysql
+
+# revision identifiers, used by Alembic.
+revision = "7def990b852e"
+down_revision = "1758275763"
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    op.add_column(
+        "oauthclient_remotetoken",
+        sa.Column(
+            "refresh_token", sqlalchemy_utils.StringEncryptedType(), nullable=True
+        ),
+    )
+    op.add_column(
+        "oauthclient_remotetoken",
+        # We are using `created` and `updated` so it should be `expires` and not `expires_at`
+        sa.Column(
+            "expires",
+            sa.DateTime().with_variant(mysql.DATETIME(fsp=6), "mysql"),
+            nullable=True,
+        ),
+    )
+
+
+def downgrade():
+    """Downgrade database."""
+    op.drop_column("oauthclient_remotetoken", "expires")
+    op.drop_column("oauthclient_remotetoken", "refresh_token")
--- a/invenio_oauthclient/handlers/refresh.py
+++ b/invenio_oauthclient/handlers/refresh.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2024 CESNET z.s.p.o.
+# Copyright (C) 2025 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Handler for refreshing access token."""
+
+from flask_oauthlib.client import OAuthResponse
+from flask_oauthlib.utils import to_bytes
+
+from invenio_oauthclient.handlers.token import make_expiration_time
+
+from ..models import RemoteToken
+from ..proxies import current_oauthclient
+
+
+def refresh_access_token(token: RemoteToken):
+    """
+    Internal method to refresh the access token (via RFC 6749 Section 6).
+
+    :param token: the remote token to be refreshed
+    :returns tuple of (access_token, refresh_token, expires)
+
+    Note: the current access/refresh token are invalidated during this call
+    """
+    remote_account = token.remote_account
+    client_id = remote_account.client_id
+    # Find the remote by client ID
+    remote = next(
+        x
+        for x in current_oauthclient.oauth.remote_apps.values()
+        if x.consumer_key == client_id
+    )
+    client = remote.make_client()
+    request_url, request_headers, request_body = client.prepare_refresh_token_request(
+        remote.access_token_url,
+        refresh_token=token.refresh_token,
+        client_id=remote.consumer_key,
+        client_secret=remote.consumer_secret,
+    )
+    resp, content = remote.http_request(
+        request_url,
+        request_headers,
+        data=to_bytes(request_body, remote.encoding),
+        method="POST",
+    )
+    resp = OAuthResponse(resp, content, remote.content_type)
+    return (
+        resp.data.get("access_token"),
+        # As per the RFC, the server MAY issue a new refresh token of an identical scope, which we must save.
+        resp.data.get("refresh_token"),
+        make_expiration_time(resp.data.get("expires_in")),
+    )
--- a/invenio_oauthclient/handlers/token.py
+++ b/invenio_oauthclient/handlers/token.py
@@ -1,13 +1,13 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2023 CERN.
+# Copyright (C) 2023-2025 CERN.
 #
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """Funcs to manage tokens."""
-
+from datetime import datetime, timedelta, timezone
 from functools import partial
 
 from flask import current_app, session
@@ -113,10 +113,31 @@ def oauth2_token_setter(remote, resp, token_type="", extra_data=None):
         secret="",
         token_type=token_type,
         extra_data=extra_data,
+        refresh_token=resp.get("refresh_token"),
+        expires=make_expiration_time(resp.get("expires_in")),
     )
 
 
-def token_setter(remote, token, secret="", token_type="", extra_data=None, user=None):
+def make_expiration_time(expires_in):
+    """Make expiration time from expires_in.
+
+    :param expires_in: The time in seconds.
+    """
+    if expires_in is None:
+        return None
+    return datetime.now(tz=timezone.utc) + timedelta(seconds=expires_in)
+
+
+def token_setter(
+    remote,
+    token,
+    secret="",
+    token_type="",
+    extra_data=None,
+    user=None,
+    refresh_token=None,
+    expires=None,
+):
     """Set token for user.
 
     :param remote: The remote application.
@@ -128,7 +149,12 @@ def token_setter(remote, token, secret="", token_type="", extra_data=None, user=
     :returns: A :class:`invenio_oauthclient.models.RemoteToken` instance or
         ``None``.
     """
-    session[token_session_key(remote.name)] = (token, secret)
+    session[token_session_key(remote.name)] = (
+        token,
+        secret,
+        refresh_token,
+        expires.isoformat() if expires else None,
+    )
     user = user or current_user
 
     # Save token if user is not anonymous (user exists but can be not active at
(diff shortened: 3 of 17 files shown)


invenio_oauthclient/handlers/token.py
```diff
@@ -112,10 +112,31 @@ def oauth2_token_setter(remote, resp, token_type="", extra_data=None):
...
 
-def token_setter(remote, token, secret="", token_type="", extra_data=None, user=None):
+def make_expiration_time(expires_in):
+    """Make expiration time from expires_in.
+
+    :param expires_in: The time in seconds.
+    """
+    if expires_in is None:
+        return None
+    return datetime.datetime.now() + datetime.timedelta(seconds=expires_in)
```
Samk13, line 127: Question: Why not use `datetime.now(timezone.utc)` instead of `datetime.now()` to ensure all datetime objects are timezone-aware for better [consistency](https://github.com/search?q=repo%3Ainveniosoftware%2Finvenio-oauthclient+now%28&type=code) and clarity?
(`utcnow()` will be deprecated in 3.12)

invenio_oauthclient/models.py
```diff
@@ -141,18 +162,37 @@ def token(self):
...
 
+    def refresh_access_token(self):
+        """Refresh the access token."""
+        if not self.refresh_token:
+            raise ValueError("No refresh token available")
+        from .handlers.refresh import refresh_access_token
+
+        access_token, refresh_token, secret, expires_at = refresh_access_token(self)
+        self.update_token(access_token, refresh_token, secret, expires_at)
+        db.session.commit()
```
Samk13, line 194: Question:
Should we use unit of work here? See: [docs](https://inveniordm.docs.cern.ch/develop/topics/uow/#unit-of-work)

mesemus: It would make invenio-oauthclient dependent on invenio-records-resources (where the UnitOfWork is defined). I would personally stick with simple commit as the rest of the library uses that, but would document it in the pydoc to make sure that the caller knows that commit will occur - does it make sense?

Samk13: Valid point, but I could consider using `db.session.begin_nested()` to allow for partial rollbacks on errors and maintain consistency with the [existing codebase](https://github.com/search?q=repo%3Ainveniosoftware%2Finvenio-oauthclient+db.session.begin_nested&type=code). 
Do you think this is a better approach?

invenio_oauthclient/handlers/token.py
Samk13: Hmm, if you changed `sa.Column("expires_at", sa.DateTime(timezone=True), nullable=True)` and SQLAlchemy still removes the timezone information despite the change, the problem could be relying somewhere else right? 
WDYT @slint?

invenio_oauthclient/models.py
mesemus: Definitely better, but I still have a small issue with that :) The problem is that when the oauth token endpoint is called, it invalidates the previous access & refresh token and returns a new pair (that is at least the case for our perun aai implementation). Then, if the new access & refresh is not stored to database (for example, the refresh method is invoked from an external begin_nested which rolls back afterwards for some reason), the original values stored in remote_token are completely unusable and any subsequent call will fail. That's why I would rather commit the token as soon as possible, or if using the begin_nested I would at least document that the caller should commit it as soon as possible - what would you prefer?

invenio_oauthclient/handlers/token.py
mesemus: Pushed the "aware" version just to see the test results for different databases.

It seems that both sqlite and mysql (https://dev.mysql.com/doc/refman/8.0/en/datetime.html) convert timestamp to utc before storing to db and then to local timezone upon retrieval - but "awareness" is lost in sqlalchemy.
Only postgres stores the timezone with the datetime ("timestamp with time zone" type) and subsequently sqlalchemy returns timezone aware datetime.

As invenio sources use utcnow() in many places (grep gave me more that 100) and even the maintrunk of sqlalchemy_utils uses utcnow(), I'd vote for keeping the datetime naive but in utc, using utcnow in the code. Later on when the invenio codebase will be upgraded to python 3.12++, then if not already solved by sqlalchemy itself we might introduce our own timestamp decorator in invenio-db (similar to https://mike.depalatis.net/blog/sqlalchemy-timestamps.html)

@Samk13 @slint What do you think?

Samk13: Switching to `datetime.now(timezone.utc)` for consistency and future-proofing is logical to me. If `sa.DateTime(timezone=True)` strips timezone info, the issue is deeper. 
I would vote to stick with `utcnow()` for now and address this in Python 3.12 upgrades. 
@slint, I recommend this. Your call. Thoughts?

invenio_oauthclient/models.py
Samk13: I see your point, it will just introduce complexity regarding token management and state consistency. 
Let's commit the token immediately after the update to ensure it's saved, avoiding potential issues with nested transactions and token invalidation. This keeps token management straightforward. 
would you agree on this?

palkerecsenyi (commented):
Thank you so much for creating this PR @mesemus, it looks like quite a bit of work and will improve security a lot while also adding support for more modern and spec-compliant OAuth servers. I'm aware it's been over a year, I'm coming back to this as I need refresh token support for the upcoming GitLab integration (https://github.com/inveniosoftware/invenio-github/pull/190) and this PR is 99% of the way there. I apologise for the large number of suggestions, these are mostly just nitpicks :)

I'm also very happy to fix them myself in a new PR if you're busy. Thanks!

invenio_oauthclient/models.py
```diff
@@ -130,6 +140,16 @@ class RemoteToken(db.Model, Timestamp):
...
 
+    @property
+    def is_expired(self):
+        """Check if access token has expired."""
+        if not self.expires_at:
+            return False
+
+        leeway = current_app.config.get("OAUTHCLIENT_TOKEN_EXPIRES_LEEWAY", 10)
+        expiration_with_leeway = self.expires_at - timedelta(seconds=leeway)
+        return expiration_with_leeway < datetime.utcnow()
```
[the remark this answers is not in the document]
palkerecsenyi: Ah I think I read this completely wrong! I see now the leeway is _reducing_ the validity rather than increasing it, so that's fully valid! And actually makes a lot of sense :)

invenio_oauthclient/models.py
```diff
@@ -129,6 +131,14 @@ class RemoteToken(db.Model, Timestamp):
     )
     """Access token to remote application."""
 
+    refresh_token = db.Column(
+        StringEncryptedType(type_in=db.Text, key=_secret_key), nullable=True
+    )
+    """Refresh token to remote application."""
+
+    expires = db.Column(db.DateTime, nullable=True)
```
ntarocco: We use a slightly different form in [invenio-records](https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/models.py#L30), can you please have a look?

invenio_oauthclient/models.py
```diff
@@ -141,18 +161,37 @@ def token(self):
...
+            or self.expiration != expires_at
+        ):
             with db.session.begin_nested():
                 self.access_token = token
                 self.secret = secret
+                self.refresh_token = refresh_token
+                self.expires_at = expires_at
                 db.session.add(self)
 
+    def refresh_access_token(self):
```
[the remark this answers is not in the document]
ntarocco: You marked this comment as resolved, but I had the same question :)
What is the answer? ☺️

invenio_oauthclient/handlers/refresh.py
```diff
@@ -0,0 +1,56 @@
...
+        request_headers,
+        data=to_bytes(request_body, remote.encoding),
+        method="POST",
+    )
+    resp = OAuthResponse(resp, content, remote.content_type)
+    return (
+        resp.data.get("access_token"),
+        # As per the RFC, the server MAY issue a new refresh token of an identical scope, which we must save.
+        resp.data.get("refresh_token"),
+        make_expiration_time(resp.data.get("expires_in")),
```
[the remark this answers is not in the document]
palkerecsenyi: As per the RFC, [`expires_in`](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1) is actually the number of seconds from the current time in which the access token will expire. E.g. `expires_in=3600` means it will expire in 1 hour's time. So this is time zone agnostic

ntarocco (approved):

slint (commented):
LGTU (with @palkerecsenyi), some extra warnings/guidance for the `db.session.commit()` and we need to update copyright headers.

invenio_oauthclient/handlers/token.py
```diff
@@ -177,7 +210,16 @@ def token_getter(remote, token=""):
         # Store token and secret in session
         session[session_key] = remote_token.token()
 
-    return session.get(session_key, None)
+    values = session.get(session_key, None)
+    if values:
+        # Continue supporting the old tuple for backwards-compatibility with existing sessions
+        if len(values) == 2:
```
kpsherva, line 216: does this mean we plan to support old style tokens indefinitely or we plan to phase them out at some point?

palkerecsenyi: We phase them out immediately. When the user logs in with OAuth and returns to the `authorized` endpoint, we check if we need extra sign up info (which is usually the case on Zenodo). If so, we need to store their credentials in the session until they complete their account, which is when we do the token exchange with the identity provider. The only case where the old token would be returned from the session is if a user was in the middle of completing the extra info form when we did the deploy, and here we are ensuring that this edge case is handled gracefully.

invenio_oauthclient/alembic/7def990b852e_add_expires_at_and_refresh_token_to_.py
```diff
@@ -0,0 +1,44 @@
+#
+# This file is part of Invenio.
+# Copyright (C) 2016-2018 CERN.
```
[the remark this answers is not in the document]
palkerecsenyi: All checked :)

invenio_oauthclient/models.py
```diff
@@ -151,18 +178,43 @@ def token(self):
...
+        ):
             with db.session.begin_nested():
                 self.access_token = token
                 self.secret = secret
+                self.refresh_token = refresh_token
+                self.expires = expires
                 db.session.add(self)
 
+    def refresh_access_token(self):
+        """Refresh the access token."""
```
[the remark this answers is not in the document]
palkerecsenyi: I've added a clearer warning

PR #335, opened 2024-09-16
views: fix boolean logic for checking remote app visibility

max-moser:
Previously, the logic did the opposite of what it was supposed to do: It filtered for the *invisible* remote apps


rekt-hard (approved):
Looks good. Thank you for catching that

PR #343, opened 2024-12-02
utils: improve group create or update

jrcastro2:


invenio_oauthclient/handlers/utils.py
```diff
@@ -113,39 +113,44 @@ def create_or_update_roles(groups):
...
+                        name=group["name"],
+                        description=group.get("description"),
+                        is_managed=False,
+                    )
+                    role = current_datastore.update_role(role_to_update)
+                    roles_ids.add(role.id)
+                else:
+                    roles_ids.add(existing_role.id)
+                
+                db.session.flush()  # Ensure changes are written before committing
```
zzacharo, line 153: Isn't that happening eitherway on `exit` of `db.session.begin_nested` ?

ntarocco: [yes](https://docs.sqlalchemy.org/en/20/orm/session_transaction.html):

> When the context manager yielded by [Session.begin_nested()](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.begin_nested) completes, it “commits” the savepoint, which includes the usual behavior of flushing all pending state. When an error is raised, the savepoint is rolled back and the state of the [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) local to the objects that were changed is expired.

I am wondering if we should delegate to the caller to call this method inside the `begin_nested`, WDYT?

jrcastro2: I thought about this, however kept it here as it was the same behaviour as before. I will change it to be the caller 👍

PR #360, opened 2025-09-22
perf(models): change `extra_data` to `JSONB`

palkerecsenyi:
Related (vaguely) to https://github.com/inveniosoftware/invenio-github/pull/190, but is not a dependency.

**Important**: we are aiming to release this at the same time as https://github.com/inveniosoftware/invenio-oauthclient/pull/328. When we merge, we need to align the DB migrations, but still keep them as 2 separate migrations so this one can be run manually. The other migration does not need to be run manually in any scenario.

---

* `JSONB` is a newer and more efficient format for storing JSON in PostgreSQL. Compared to `JSON`, it offers significantly faster querying and indexing support (but is very slightly slower to input). The JSON is decomposed into a binary format rather than being stored in plaintext.

* From a high-level perspective, the access of the data is unchanged in Python so this is not a breaking change. However, the new column type supports more querying operations which could be used in the future.

* Migrating the column type is quite easy for small tables but can cause performance and stability issues for larger ones. Therefore, a default Alembic migration has been included for the vast majority of use cases, as well as a step-by-step alternative guide for larger instances. Instances with more than ~50k rows in `oauthclient_remoteaccount` are advised not to use the Alembic migration and to instead follow the upgrade guide.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,12 @@
 Changes
 =======
 
+Version v6.0.0 (released 2025-XX-XX)
+
+- perf(models): changed SQL column type of `extra_data` to `JSONB` instead of `JSON` for PostgreSQL
+
+  - This migration could flood the WAL when run on very large `oauthclient_remoteaccount` tables (~50k+ rows) as well as causing a full lock that lasts several minutes and stops users from being able to log in with third-party accounts. This won't happen in the majority of cases. For instances not meeting this size criteria, the automated Alembic migration will suffice. However, for tables above this size, please use instead the step-by-step instructions in the upgrade guide (included in the documentation of this module).
+
 Version v5.3.1 (released 2025-10-21)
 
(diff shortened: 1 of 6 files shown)


ntarocco (approved):

palkerecsenyi:
The following PRs propagate the major version bump:

- https://github.com/inveniosoftware/invenio-app-rdm/pull/3226
- https://github.com/inveniosoftware/invenio-users-resources/pull/186
- https://github.com/inveniosoftware/invenio-github/pull/201
- https://github.com/inveniosoftware/invenio-rdm-records/pull/2193
- https://github.com/inveniosoftware/invenio-app-ils/pull/1252
- https://github.com/CERNDocumentServer/invenio-cern-sync/pull/11

The relevant docs for this and #328 are in https://github.com/inveniosoftware/docs-invenio-rdm/pull/884

PR #372, opened 2026-01-30
fix(logout): remove unmanage_roles from session

### LINKED ISSUE #3156: User impersonation doesn't work as intended
Package version (if known): v13.0.0b4.dev2

## Describe the bug
When impersonating a user with no admin role, you can still access the admin panel and edit every record.

## Expected behavior
You should be able to access exactly what the user can.

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/3156

### DIFF
--- a/invenio_oauthclient/handlers/token.py
+++ b/invenio_oauthclient/handlers/token.py
@@ -235,10 +235,20 @@ def token_delete(remote, token=""):
 
 
 def oauth_logout_handler(sender_app, user=None):
-    """Remove all access tokens from session on logout."""
+    """Remove all access tokens and OAuth session data on logout.
+
+    This includes:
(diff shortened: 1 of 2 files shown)


egabancho (approved):

egabancho:
Just wondering, doesn't the session get cleaned after logout?

zzacharo:
> Just wondering, doesn't the session get cleaned after logout?

Flask-login [deletes](https://github.com/maxcountryman/flask-login/blob/c8bba84b9ba6768e878317fc46c54bd13fa1ac07/src/flask_login/utils.py#L204) only specific keys from the session. The `user_logged_out ` signal handles the rest.

PR #384, opened 2026-07-21
fix(tests): remove group handler tests

max-moser:
* since recent releases of `Invenio-Accounts`, roles must have the same value for their ID and name
* the removed tests were checking for some functionality that set the ID of a role to something different from its name
* since this is not possible anymore, these tests became obsolete and subject to removal


max-moser:
closed because something was changed in invenio-accounts that caused this failure

